PR #3, opened 2022-07-15
Feature: pair plots

jobrachem:
Add the functions `gs.plot_scatter` and `gs.plot_pairs`.

Note that I also changed how parameter indices are displayed in plot labels. They are displayed now using square brackets instead of underscores. I think this provides a clearer distinction between the parameter name, which may contain underscores, and the index used to access a specific subparameter.

Before: beta_0
After: beta[0]

You can use this notebook for checking out the new functionality: [202207150909-pairplot.ipynb.zip](https://github.com/liesel-devs/liesel/files/9118973/202207150909-pairplot.ipynb.zip)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -85,7 +85,7 @@ ipython_config.py
 # pyenv
 #   For a library or package, you might want to ignore these files since the code is
 #   intended to run in multiple environments; otherwise, check them in:
-# .python-version
+.python-version
 
 # pipenv
 #   According to pypa/pipenv#598, it is recommended to include Pipfile.lock in version control.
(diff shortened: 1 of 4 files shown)


wiep (commented):
very nice.

If i'm correct, you changed the interface slightly by renaming one column in the data frame used to create the plots `param` -> `param_label`. This requires a change in the existing functions default values. Could you explain why you did that?

jobrachem:
You're totally right. I should have mentioned that explicitly. 
Previously, the plot data were curated such that there was a column `param`, which contained both the name of the parameter and the parameter index:

image

This is untidy data, because two kinds of information are combined in a single column (the parameter name and the index). I changed this behavior such that we retain the original param name, the param index, and also have the label for the plot at hand:

image

This change is not strictly necessary, I just came across and thought in the future it may come in handy.

wiep (approved):
ok great. i think we can have the api change. i think it will not affect many already implemented use cases (if any at all).

PR #7, opened 2022-08-16
Add mapping from position name to kernel identifier to SamplingResults

jobrachem:
Additionally to what's mentioned in the title, this PR also streamlines the naming of `SamplingResults`, bringing everything to the plural form. The singular form is still available for backwards compatibility.

I have not added a printout of the kernel identifier to the `gs.Summary.__repr__` at this stage. Do you want to have it in this PR @wiep?


wiep:
the name in the table would be pretty nice i think. however, then the table would be even broader. so i would not do it for now. @hriebl, would you like to have it?

hriebl:
I think an extra column in the summary table would be nice. liesel-devs/liesel-internal#109 also suggests that we want this. Maybe we can reduce the width of the summary table by removing some other column?

jobrachem:
@hriebl since Paul is on vacation now for a bit, I think it may be best if we decide how to proceed. Since you both expressed interest in having the extra column, I propose that I add the column – it should be easy to remove another column later on, if we wish to do so. Would you agree to this?

hriebl:
I agree. It should be straightforward to add the column, see [the `to_dataframe()`](https://github.com/liesel-devs/liesel/blob/main/liesel/goose/summary_m.py#L662) and [`_param_df()` methods](https://github.com/liesel-devs/liesel/blob/main/liesel/goose/summary_m.py#L729) of the `Summary` object. When you're done, we can still decide if some other column needs to be removed from the table.

jobrachem:
The column is now included, as per https://github.com/liesel-devs/liesel/pull/7/commits/5bf4d91f5765f1783bcaababc9dc0f2a119dc1f6. 

Here's a notebook for testing, if you wish:
[110-summary.ipynb.zip](https://github.com/liesel-devs/liesel/files/9373450/110-summary.ipynb.zip)

In this example, the output looks like this:

image

jobrachem:
@hriebl I implemented the changes we talked about :)

hriebl:
Thank you for the update! Merging.

PR #8, opened 2022-08-24
Bump jax and jaxlib versions

jobrachem:
Closes https://github.com/liesel-devs/liesel-internal/issues/154

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -19,8 +19,8 @@ python_requires = >=3.10
 install_requires =
     arviz>=0.11
     blackjax>=0.8.2,<0.9
-    jax==0.3.13
-    jaxlib==0.3.10
+    jax==0.3.16
+    jaxlib==0.3.15
     jaxopt>=0.4,<0.5
     matplotlib>=3.5
     networkx>=2.6


jobrachem:
Hey @hriebl, I think this is ready for review. The tests seem to pass just fine.

jobrachem:
Converted this back to a draft, because the version bump messes up our logging.

1. The `"absl"` logger spams the output with all kinds of messages, starting form `DEBUG` level. This can be avoided by explicitly setting its level to a higher level.
2. Liesel's own logging messages are duplicated, apparently by another StreamHandler with the default format. Example:

```
INFO:liesel.goose.engine:Starting epoch: FAST_ADAPTATION, 75 transitions, 25 jitted together
INFO - Starting epoch: FAST_ADAPTATION, 75 transitions, 25 jitted together
```

This is the same message, printed twice with the same format.

jobrachem:
As it turns out, this issue was caused by the fact that I was using `tfd.Normal(0, 1)` as a default argument value in my CTM code, causing some code executing during the import of `liesel_ctm`. 

I do not see any more issues with bumping the jax and jaxlib versions. @hriebl, this PR is ready I think :)

PR #12, opened 2022-09-15
Provide basic docstring decorator; update goose docstrings

jobrachem:
A basic docstring decorator that only directly reuses docstrings. A more sophisticated version may follow at a later point.

Can be tested and used as follows:

```python
from liesel.docs import usedocs

class ParentClass:
    '''Parentclass documentation'''

    def parent_method(self):
        '''Method documentation.'''

@usedocs(ParentClass)
class InheritingClass:

    @usedocs(ParentClass.parent_method)
    def parent_method(self):
        pass


print(InheritingClass.__doc__)
print(InheritingClass.parent_method.__doc__)
```


liesel/docs.py
```diff
@@ -0,0 +1,36 @@
+from functools import wraps
+
+
+def usedocs(of: object):
+    """
+    Replaces the docstring of the decorated object with the docstring of the specified
+    object.
+
+    In classes, methods need to be decorated individually. Attribute docstrings are not
+    covered by this decorator.
```
hriebl, line 10: Would it make sense to add a decorator that fills all empty method docstrings of a class with the method docstrings from another class? Maybe that would be useful for the kernel protocols and implementations?

liesel/docs.py
```diff
@@ -0,0 +1,36 @@
...
+
+    Usage example::
+
+        class ParentClass:
+            '''Parentclass documentation'''
```
hriebl, line 15:
```suggestion
            '''Parent class documentation.'''
```

liesel/docs.py
```diff
@@ -0,0 +1,36 @@
...
+
+        @usedocs(ParentClass)
+        class InheritingClass:
+
+            @usedocs(ParentClass.parent_method)
+            def parent_method(self):
+                pass
+    """
+
+    def _use_docs(obj):
```
hriebl, line 28:
```suggestion
    def _usedocs(obj):
```

For consistency?

hriebl (approved):
Just a few minor comments. If you would like to keep things as they are, I'm also fine and ready to merge this.

jobrachem:
@hriebl The decorator is now, as I call it, greedy by default. That means, it will also try to copy method and property docstrings. This can be turned off by using the argument `greedy=False`.

Also, the decorator will never overwrite existing docstrings.

jobrachem:
@GianmarcoCallegher  I converted back to a draft. I'll message you probably tomorrow when the review can actually start.

wiep:
@GianmarcoCallegher, please review and merge (with a squash).

hriebl:
I made a few cosmetic changes. IMO this is ready to be merged.

PR #15, opened 2022-10-23
Make Sphinx respect the module __all__ attribute

hriebl:
See [the Sphinx docs](https://www.sphinx-doc.org/en/master/usage/extensions/autosummary.html#confval-autosummary_ignore_module_a) and sphinx-doc/sphinx#9831. Please review, @jobrachem! :eyes:

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -123,6 +123,9 @@ autosummary_filename_map = {
     "liesel.goose.summary_m.Summary": "liesel.goose.summary_m.summary-class",
 }
 
+# Only document module members that are listed in __all__ (if defined).
+autosummary_ignore_module_all = False
+
 # Remove auto-generated API docs from the sidebar. They take too long to build.
 remove_from_toctrees = [
     "generated/liesel.liesel.*.*.*.rst",


jobrachem:
Looks sensible to me. This is a nice additional way for us to control which objects we want to have documented. The docs build without problems on my machine.

wiep:
isn't the more pythonic way to make them private with an `_` prefix?

jobrachem:
You have a point there, of course. The `__all__` attribute mainly solves an issue arising with wildcard imports ([docs](https://docs.python.org/3/tutorial/modules.html#importing-from-a-package)). It is also referred to as "an explicit index of the package". If we use it as the explicit definition of the public API, we get three sorts of objects: 1) Explicitly public, 2) Neither explicitly public, nor explicitly private, 3) explicitly private. I am not sure about what exactly would go into the second category, and that of course may be a problem. But it could also be empty, and then we just have an explicit list of all public objects of the class, which may be nice to have.

I guess it comes down to our judgment of whether we want `__all__` to be a place where we explicitly define our public API. From the library code that I saw so far, I inferred that this was the case, but I guess I was mistaken there. Would you like to discuss this more thoroughly on thursday @wiep?

wiep:
I don't necessarily believe that we should make everything `_` aka private if we want to avoid documentation. I would usually say, that one wants to use reexports to define an public api. The public docu considers only those modules while the private/internal documentation considers all modules.

Maybe using `__all__` is a good intermediate approach or alternative approach. 

Can sphinx handle two different entry point to create public and internal documentation?

jobrachem:
> I don't necessarily believe that we should make everything `_` aka private if we want to avoid documentation. I would usually say, that one wants to use reexports to define an public api. The public docu considers only those modules while the private/internal documentation considers all modules.
>
> Maybe using `__all__` is a good intermediate approach or alternative approach.

I think using the `__all__` attribute is actually the best way to make sphinx document only the right classes. Objects imported in the `__init__.py` (= reexported I think) and members listed in `__all__` are actually the same in liesel – and the pre-commit will fail an object is imported in the `__init__.py` and not listed in `__all__` – at least if the object is not used in another way in the init.

> Can sphinx handle two different entry point to create public and internal documentation?

My first guess is that I do not think it does.

wiep:
One can also re-export in non-init modules to have an hierarchical api. Anyway, if `__all__` is the most common way to define an public interface, i'm fine with using it.

> > Can sphinx handle two different entry point to create public and internal documentation?
> 
> My first guess is that I do not think it does.

maybe in two different folders, `/docs`, `/docs-internal` in which different templates are used.

jobrachem:
>One can also re-export in non-init modules to have an hierarchical api. Anyway, if __all__ is the most common way to define an public interface, i'm fine with using it.

I see 👍 Cool. I'm not sure if it is the most common one - but it seems quite legit to me. Sphinx even thinks about making the reliance on `__all__` the default. But if we notice any problems with this, thankfully it is easy to change :)

> maybe in two different folders, `/docs`, `/docs-internal` in which different templates are used.

That would surely be a possibility. But that would mean having two separate documentations. I would be quite reluctant to set up such a system. Maybe it makes sense to talk in person about this

wiep:
i don't think that we really need it. our codebase of private functions/classes is rather small.

PR #16, opened 2022-10-25
Update versions of pre-commit-hooks

wiep:
the main change -- and the reason why so many files have been modified -- is that the new version of `isort` replaces a lot of imports from `typing` with imports from `collections.abc` (see https://github.com/asottile/pyupgrade/issues/558). Importing from `typing` is supposedly deprecated since 3.9.


hriebl (approved):
Some housekeeping, great, thank you!

PR #19, opened 2022-12-14
updates PyMC requirement

wiep:
Set's pymc requirement to >= 5.0.0.

Also, fixes code to configure pymc's/pytensor's default float type to follow PyMC's change of the dependency from aesera to pytensor which is a fork of aesera.


wiep:
@hriebl, this PR bumps the jax, laxlib requirements to 0.4.1.

jobrachem:
I think it would be nice to make a minor release quite soon, once this PR is merged. Otherwise, the liesel installation from pypi is broken.

hriebl:
@jobrachem, why is it currently broken?

hriebl:
Hmm, I don't seem to get any import errors on my system.

The fact that the tests fail for 9ea3700ae5e3bf6fdf907654c522c213b5c4db3f seems to be unrelated to this specific commit or PR. It might have something to do with the new NumPy release 1.24 (`pip install numpy==1.23.5` partially solves the problem) and the fact that `Summary.to_dataframe()` returns columns of dtype object in some circumstances. I haven't been able to figure out more, though.

Could any of you investigate this further?

jobrachem:
> There's an import error because of (I think) the new Jax version that gets installed alongside Liesel in a fresh venv. Will test this again and post the full error message in the morning, am currently only on the phone.

I cannot reproduce the issue that I had yesterday in a fresh virtualenv today. Yesterday, I was on an *existing* virtualenv and my steps, as I recall them, were 

1. `pip uninstall liesel`, 
2. `pip install liesel`, 
3. Try to `import liesel.model as lsl` and `import liesel.goose as gs`
4. Observe importerror.
5. Downgrade jax and jaxlib to the latest `0.3.x` versions.
6. Try again, now it works.

But since I cannot reproduce the issue today, I can only conclude that my message was premature.

jobrachem:
> The fact that the tests fail for [9ea3700](https://github.com/liesel-devs/liesel/commit/9ea3700ae5e3bf6fdf907654c522c213b5c4db3f) seems to be unrelated to this specific commit or PR. It might have something to do with the new NumPy release 1.24 (`pip install numpy==1.23.5` partially solves the problem) and the fact that `Summary.to_dataframe()` returns columns of dtype object in some circumstances. I haven't been able to figure out more, though.
> 
> Could any of you investigate this further?

This is indeed the same error as the one we had in https://github.com/liesel-devs/liesel-internal/issues/200#issuecomment-1358816001. I cannot pin down the issue so far though. The data types in the plot_df columns don't look too suspicious to me:

```
plot_df.info()
<class 'pandas.core.frame.DataFrame'>
RangeIndex: 30000 entries, 0 to 29999
Data columns (total 6 columns):
 #   Column       Non-Null Count  Dtype   
---  ------       --------------  -----   
 0   param        30000 non-null  object  
 1   param_index  30000 non-null  int64   
 2   chain_index  30000 non-null  category
 3   iteration    30000 non-null  int64   
 4   value        30000 non-null  float32 
 5   param_label  30000 non-null  object  
dtypes: category(1), float32(1), int64(2), object(2)
memory usage: 1.1+ MB
```

hriebl:
Thank you for pointing me to that issue! I had missed it while in Florence.

hriebl:
I think I found the reason for the failing tests on this PR:

- https://github.com/mwaskom/seaborn/issues/3192
- https://github.com/numpy/numpy/issues/22826

hriebl:
I followed https://github.com/mwaskom/seaborn/pull/3194 and marked NumPy 1.24.0 as incompatible in this PR.

hriebl (approved):
Now that the tests are working, I'm happy with this PR. Should we go ahead and merge?

PR #23, opened 2023-02-08
Remove sets, make dist_reg_mcmc() deterministic, closes #20

### LINKED ISSUE #20: The order of the MCMC kernels from dist_reg_mcmc() is non-deterministic
The order of the MCMC kernels from `dist_reg_mcmc()` is non-deterministic, which makes it impossible to generate reproducible MCMC chains with this function. For the tests and the tutorials, it would be great to have reproducible MCMC chains.

The non-deterministic behavior comes from [this loop](https://github.com/liesel-devs/liesel/blob/v0.2.1/liesel/model/distreg.py#L307), and originally from the fact that the `DistRegBuilder` and the `GraphBuilder` use [sets to manage their nodes and variables](https://github.com/liesel-devs/liesel/blob/v0.2.1/liesel/model/model.py#L111-L115).

The easy solution would be to sort the groups in the `dist_reg_mcmc()` function by the name of one of the contained nodes or variables. If we want to solve the problem more fundamentally, we could also investigate how to make the `GraphBuilder` deterministic as a whole.

hriebl:
See title and #20.


liesel/model/model.py
```diff
@@ -235,12 +235,12 @@ def add(self, *args: Node | Var | GraphBuilder) -> GraphBuilder:
...
             elif isinstance(arg, Var):
-                self.vars.add(arg)
+                self.vars.append(arg)
             elif isinstance(arg, GraphBuilder):
-                self.nodes.update(arg.nodes)
-                self.vars.update(arg.vars)
+                self.nodes.extend(arg.nodes)
+                self.vars.extend(arg.vars)
             else:
                 raise RuntimeError(f"Cannot add {type(arg).__name__} to graph builder")
```
wiep, line 245: The set has no duplicates but the list could hold the same node twice. Do we assume unique elements somewhere else in the code so that we should check for that here?

wiep (commented):
See comment.

liesel/model/model.py
hriebl: Uniqueness is ensured [in the `_all_nodes_and_vars()` method](https://github.com/liesel-devs/liesel/blob/deterministic-distreg-mcmc/liesel/model/model.py#L176-L178), which is used for all relevant operations (at least, I hope so)...

wiep: i'm not sure if the function does it. consider two different nodes which have the same input node, the input node will be twice in the node list.

see line 198f.
> nodes.extend(node.all_input_nodes())

hriebl: It does. We are filling the `all_nodes` list here, `nodes` is only the list we're iterating over:

``` python
while nodes:
    node = nodes.pop()

    if node in all_nodes:
        continue

    nodes.extend(node.all_input_nodes())
    all_nodes.append(node)
```

The relevant line is `if node in all_nodes: ...`.

wiep (approved):

hriebl:
I removed almost all sets from the nodes module, now the order of the kernels should really be deterministic. Please take another quick look, @wiep!

liesel/model/nodes.py
```diff
@@ -45,6 +46,12 @@
 Distribution = Union[jd.Distribution, nd.Distribution]
 Bijector = Union[jb.Bijector, nb.Bijector]
 
+T = TypeVar("T")
+
+
+def _unique_tuple(*args: Iterable[T]) -> tuple[T, ...]:
+    return tuple(dict.fromkeys(chain(*args)))
+
```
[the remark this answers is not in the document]
wiep: is resolved in 8fd69b043a0a6108db382f4233416d2f86d411d9.

PR #24, opened 2023-02-08
Add mutable group object

jobrachem:
This PR holds my proposal for a `lsl.Group` object.

## How to use the `Group` object

Add two vars to a group upon initialization::

```python

>>> v1 = lsl.Var(0.0, name="v1")
>>> v2 = lsl.Var(0.0, name="v2")
>>> group = lsl.Group("g1", var1=v1, var2=v2)
>>> group.vars
mappingproxy({'var1': Var<v1>, 'var2': Var<v2>})
``` 

Add a var to a group after initialization::

```python

>>> v1 = lsl.Var(0.0, name="v1")
>>> group = lsl.Group("g1")
>>> group["var1"] = v1
>>> group.vars
mappingproxy({'var1': Var<v1>})
```

Remove var from a group. This ends the group membership of the variable, i.e.: the var is removed from the group AND the group is removed from the var::

```python

>>> v1 = lsl.Var(0.0, name="v1")
>>> group = lsl.Group("g1", var1=v1)

>>> del group["var1"]
>>> group.vars
mappingproxy({})
>>> v1.groups
{}
```

Remove group from var. This also ends the group membership of the variable, i.e.: the var is removed from the group AND the group is removed from the var::

```python

>>> v1 = lsl.Var(0.0, name="v1")
>>> group = lsl.Group("g1", var1=v1)

>>> del v1.groups["g1"]
>>> v1.groups
{}
>>> g1.vars
mappingproxy({})
```

Create a new group that merges two groups::

```python

>>> v1 = lsl.Var(0.0, name="v1")
>>> v2 = lsl.Var(0.0, name="v2")

>>> group1 = lsl.Group("g1", var1=v1)
>>> group2 = lsl.Group("g2", var2=v2)

>>> group3 = lsl.Group("g3", **group1, **group2)
>>> group3.vars
mappingproxy({'var1': Var<v1>, 'var2': Var<v2>})
```

Merge a group into an existing group::

```python

>>> v1 = lsl.Var(0.0, name="v1")
>>> v2 = lsl.Var(0.0, name="v2")

>>> group1 = lsl.Group("g1", var1=v1)
>>> group2 = lsl.Group("g2", var2=v2)

>>> group1 |= group2
>>> group1.vars
mappingproxy({'var1': Var<v1>, 'var2': Var<v2>})
```

## Other changes

- `Node.groups` and `Var.groups` are now basically a modified dictionary of groups, indexed by their names. It is modified to prevent user errors. For example, you cannot add groups to it, if the node/var is not a member of that group. And you cannot simply override existing groups with the same name - to do that, you first have to remove the node/var from the existing node.
- The changed allowed `Model.groups()` to be simplified a bit. It (still) returns a dictionary of all groups in the model. Note that this method will use ordinary dictionary merging. That means it will not complain if there are different groups with the same name - but only one of those will be included in the dictionary in the end. This mirrors the previous behavior, I think. I did not change that, because I am not sure whether it is really important to have a check here.
- `GraphBuilder.groups()` is new and does the same as `Model.groups()`
- `GraphBuilder.add_group()` becomes `GraphBuilder.add_groups()`. It takes a variable number of groups and adds their members to the graph. Before, it created the groups, now the groups are created on their own. This method checks for name uniqueness against the existing groups obtained from `GraphBuilder.groups()`. It also checks that the entered groups have unique names.

### DIFF
--- a/liesel/model/__init__.py
+++ b/liesel/model/__init__.py
@@ -26,6 +26,7 @@ from .nodes import (  # TODO: Bijector?
     Data,
     Dist,
     Distribution,
+    Group,
     InputGroup,
     Node,
     NodeState,
--- a/liesel/model/distreg.py
+++ b/liesel/model/distreg.py
@@ -27,7 +27,7 @@ from .legacy import (
     SmoothingParam,
 )
 from .model import GraphBuilder, Model
-from .nodes import Array, Bijector, Dist, Distribution, Node, Var
+from .nodes import Array, Bijector, Dist, Distribution, Group, Node, Var
 
 matrix_rank = np.linalg.matrix_rank
 
@@ -78,7 +78,7 @@ class DistRegBuilder(GraphBuilder):
         s: float,
         predictor: str,
         name: str | None = None,
-    ) -> dict[str, Var]:
+    ) -> Group:
         """
         Adds a parametric smooth to the model builder.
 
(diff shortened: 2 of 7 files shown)


hriebl:
Closing this for now as it's not being actively worked on.

PR #25, opened 2023-02-09
Use variable instead of node names in dist_reg_mcmc(), closes #22

### LINKED ISSUE #22: dist_reg_mcmc() uses node names instead of variable names as position keys
The `dist_reg_mcmc()` function uses node names instead of variable names as position keys, which can be seen in the table and the figures in [the Metropolis-in-Gibbs section in the GEV tutorial](https://github.com/liesel-devs/liesel/blob/3c5023472a418fd0433afd1c409623135fe7fd0c/docs/source/tutorials/md/04-mcycle.md#metropolis-in-gibbs), e.g. the first position key is `loc_np0_beta_value` instead of `loc_np0_beta`.

The `GooseModel` supports using variable names as position keys, so the required changes to the `dist_reg_mcmc()` function should be minor, and the MCMC summary would feel a lot less clumsy for the user.

hriebl:
See title and #22. The code modifications are minimal.


jobrachem (approved):
Looks good. 

I have two loosely related comments. They are rather thoughts on what comes next than change requests on this PR.

Apart from these considerations, I think we can merge this PR. I'd just like to write down these thoughts here for further discussion - maybe we can talk about them tomorrow @hriebl  

### First point: Create transition functions without a group

In the `01-lin-reg.qmd`, we still need to explicitly include the `_value` name part, because there, we do not have a group:

```python
def draw_sigma(prng_key, model_state):
    a_prior = model_state["a_value"].value
    b_prior = model_state["b_value"].value
    n = len(model_state["y_value"].value)

    resid = model_state["y_value"].value - model_state["y_hat_value"].value

    a_gibbs = a_prior + n / 2
    b_gibbs = b_prior + jnp.sum(resid**2) / 2
    draw = b_gibbs / jax.random.gamma(prng_key, a_gibbs)
    return {"sigma_value": draw}
```

Maybe we could provide a little helper function that simply extracts the right value from the model state, something like this:

```python
def get_value(model_state, name):
    return model_state[name + "_value"].value
```

It's not big, but I always feel that little things like this really add up to make code much more readable, especially when the code is re-used often.

### Second point: `Group.get_value()` method

To simplify the pattern of filtering a group and then obtaining the value of a var, we could add a little method to the `Group` object (#24) 

The current pattern is:

```python
group = {k: x if isinstance(x, Node) else x.value_node for k, x in group.items()}
...
a_prior = model_state[group["a"].name].value
```

The proposed method would be

```python
class Group:
    def get_value(self, model_state, name):
        value_name = self.__getitem__(name).value_node.name
        return model_state[value_name].value
```

This would change the above pattern to

```python
a_prior = group.get_value(model_state, "a")
```

This is of course assuming that the input `group` argument of the `tau2_gibbs_kernel()` function would be of type `Group`.

PR #27, opened 2023-02-15
Keep compiled tutorials out

jobrachem:
This PR introduces a GitHub action that renders our tutorials and then triggers a documentation build on readthedocs.

@hriebl upon merging this PR, we should remove the old webhook that triggers readthedocs builds, because this action will replace that one.


hriebl (approved):
I simplified the action a bit, LGTM now.

jobrachem:
@hriebl I forgot to bring that up yesterday when you showed me your changes: I think naming the action `quarto.yml` is a bit confusing, because this action does not only render our quarto tutorials, but also takes over the whole responsibility of triggering documentation builds. That's why we also still need to disable the previous Webhook in the repository settings. So I think `docs.yml` is actually a better indication of what this action does for us.

hriebl:
Okay, sorry about that. We can change it back.

PR #28, opened 2023-02-16
make groups immutable

jobrachem:
This PR makes groups immutable. Some notes:

1. Rather than inheriting from `MappingProxyType`, the immutable group object now acts as a wrapper around a private dictionary `Group._nodes_and_vars`.
2. Groups are now only filled upon initialisation.
3. There's no intended way to remove a member from a group. Membership is forever, except if you mess with private attributes.
4. I am now using (private) dictionaries to store the groups in the nodes and variables, even though our whiteboard sketch used a private list and a property that turns the list into a dict. The reason is that I don't want to allow a situation with multiple groups of the same name being stored in the list. Upon turning that list into a dict, only one of these groups would survive. This can of course be prevented in different ways. I now check for name uniqueness upon group creation.
5. As we discussed @hriebl, I did not implement any weak referencing functionality. I am also not so sure anymore that it would be beneficial for us to have it. If the nodes and vars only hold weak references to the groups, the groups may actually be lost too early, such that the nodes and vars do not carry all model-relevant information anymore.


jobrachem:
@hriebl this is ready for you to take a look at.

liesel/model/nodes.py
```diff
@@ -976,5 +951,85 @@ def add_group(name: str, **kwargs: Node | Var) -> None:
...
+            name: obj
+            for name, obj in self._nodes_and_vars.items()
+            if isinstance(obj, Node)
+        }
+        return MappingProxyType(nodes)
+
+    @property
+    def nodes_and_vars(self) -> MappingProxyType[str, Node | Var]:
+        """A mapping of all group members with their names as keys."""
+        return MappingProxyType(self._nodes_and_vars)
```
hriebl, line 1003: These properties could as well be attributes that are filled upon initialization because of the immutability.

liesel/model/nodes.py
```diff
@@ -976,5 +951,85 @@ def add_group(name: str, **kwargs: Node | Var) -> None:
...
+        model_state
+            A goose model state.
+        name
+            Name of the variable within this group.
+
+        Returns
+        -------
+        The variable's value.
+        """
+        member = self.__getitem__(name)
```
hriebl, line 1021: Does `self[name]` not work here?

liesel/model/nodes.py
```diff
@@ -976,5 +951,85 @@ def add_group(name: str, **kwargs: Node | Var) -> None:
...
+            if isinstance(obj, Node)
+        }
+        return MappingProxyType(nodes)
+
+    @property
+    def nodes_and_vars(self) -> MappingProxyType[str, Node | Var]:
+        """A mapping of all group members with their names as keys."""
+        return MappingProxyType(self._nodes_and_vars)
+
+    def get_value(self, model_state: dict, name: str) -> Array:
```
hriebl, line 1005: Can we find a more descriptive name here? Maybe `value_from_model_state`?

liesel/model/nodes.py
hriebl, line 1003: These properties could as well be attributes that are filled upon initialization because of the immutability.

liesel/model/nodes.py
jobrachem: No, it works, we can use it.

liesel/model/nodes.py
jobrachem: I would prefer `get_value` over `value_from_model_state`, because 1) I think keeping this method name short is important 2) I think it is already quite clear from the argument name and docstring that the value is obtained from the model state.

Would you like something like `Group.value_from(model_state, name)` better?

liesel/model/nodes.py
jobrachem: Good point, I'll change that.

hriebl:
This looks really nice and simple to me. I've added a few first comments. Also, could you please resolve the conflict against main, @jobrachem? :slightly_smiling_face:

liesel/model/nodes.py
jobrachem: We are going with `.value_from` now.

hriebl (approved):
I fixed some minor issues, merging this now. Thank you very much for your work, @jobrachem!

PR #29, opened 2023-02-17
Improve uniqueness checks for group names

hriebl:
With these improved uniqueness checks, (1) it is now possible to add the same group twice to a graph builder, and (2) the group names are now checked for uniqueness when added directly to a model without a graph builder.

Do you see any problems, @jobrachem? :eyes:

### DIFF
--- a/liesel/model/model.py
+++ b/liesel/model/model.py
@@ -673,18 +673,32 @@ class Model:
             nodes_and_vars = [*model.nodes.values(), *model.vars.values()]
             model.pop_nodes_and_vars()
 
-        groups = {g for nv in nodes_and_vars for g in nv.groups.values()}
-        self._nodes = {nv.name: nv for nv in nodes_and_vars if isinstance(nv, Node)}
-        self._vars = {nv.name: nv for nv in nodes_and_vars if isinstance(nv, Var)}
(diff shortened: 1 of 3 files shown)


liesel/model/model.py
```diff
@@ -675,9 +673,13 @@ def __init__(
             nodes_and_vars = [*model.nodes.values(), *model.vars.values()]
             model.pop_nodes_and_vars()
 
+        groups = {g for nv in nodes_and_vars for g in nv.groups.values()}
         self._nodes = {nv.name: nv for nv in nodes_and_vars if isinstance(nv, Node)}
         self._vars = {nv.name: nv for nv in nodes_and_vars if isinstance(nv, Var)}
 
+        if len({g.name for g in groups}) < len(groups):
+            raise RuntimeError("Model received groups with duplicate names")
```
jobrachem, line 681: Small thing: I think the error message is more helpful, if the duplicate name shows up in there.

jobrachem (approved):
Just a minimal comment about the error message for duplicate groups, see above. Looks nice!

liesel/model/model.py
hriebl: As you wish! :rocket:

PR #30, opened 2023-02-17
Add labels to tutorial code chunks

jobrachem:
This addresses the labels-point in https://github.com/liesel-devs/liesel-internal/issues/141.

I also took the opportunity to remove some leftover compiled tutorial files and the makefile. Since now the CI is rendering the tutorials, we do not need the makefile anymore.


jobrachem:
@hriebl would you do the review?

hriebl:
Sure! Is it a good idea to have spaces in the labels? In tutorial three and four I used dashes, because I assumed that was better for the file names...

PR #35, opened 2023-02-21
Add Group docstring

jobrachem:
This adds a docstring to the group object and adds it to the `__all__` attribute in `liesel.model.nodes`.

### DIFF
--- a/liesel/model/nodes.py
+++ b/liesel/model/nodes.py
@@ -30,6 +30,7 @@ __all__ = [
     "Data",
     "Dist",
     "Distribution",
+    "Group",
     "InputGroup",
     "Node",
     "NodeState",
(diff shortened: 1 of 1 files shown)


jobrachem:
@hriebl  Would you review?

hriebl (approved):
Great explanation of the `Group` object. Thank you very much!

jobrachem:
Thanks for the proofreading @hriebl! I'm wondering about one thing: I often write something like `0.` instead of `0.0` to declare floats, because I like the brevity. Do you think that's not a good practice?

hriebl:
I absolutely don't mind, but Black prefers `0.0` over `0.`, so I usually follow that style:

``` sh
(liesel) [hannes@dell-fedora ~]$ echo "0." | black -
0.0
reformatted -

All done! ✨ 🍰 ✨
1 file reformatted.
```

PR #36, opened 2023-02-21
Allow docs to be built locally

### LINKED ISSUE #32: Allow docs to be built locally as well as on RTD
Because of the new way to render tutorials (#27) and the corresponding changes to `conf.py`, at the moment, you need to comment a few lines to be able to build the docs locally. It would be nice if we did not have to do this.

jobrachem:
Closes #32


hriebl (approved):
Works on my system. Good job, thank you!

PR #38, opened 2023-02-23
Sample from mvn degenerate distribution

### LINKED ISSUE #34: Implement sampling from MultivariateNormalDegenerate
Credit: @hriebl 

If we have a singular precision matrix $P$ of dimension $k$, we can use the eigenvalue decomposition like this:

$$
P = \Sigma^+ = Q A^+Q^T,
$$

where $Q$ is the orthogonal matrix of eigenvectors and $A^+$ is the diagonal matrix of the eigenvalues of $P$. If $P$ is singular, $A^+$ contains zero entries. Now we can take the inverse of the non-zero elements of $A^+$ to receive $A$. Then we can write $\Sigma = Q A Q^T$.

To sample from $N(0, \Sigma)$, we can first sample 

$$
X \sim N(0, I_k),
$$

where $I_k$ is the identity matrix of dimension $k$. Then we transform $X$ by applying $QA^{1/2}X$, such that

$$
\begin{aligned}
\text{Cov}(QA^{1/2}X) &= QA^{1/2}I(A^{1/2})^TQ^T\\
& = Q A Q^T\\
& = \Sigma.
\end{aligned}
$$

Thus, the generalized inverse of $\text{Cov}(QA^{1/2}X)$ is the precision matrix that we want. The implementation will require only small changes to the ecisting `MultivariateNormalDegenerate` class. We can use [jax.numpy.linalg.eigh](https://jax.readthedocs.io/en/latest/_autosummary/jax.numpy.linalg.eigh.html#jax.numpy.linalg.eigh).

jobrachem:
Closes #34.

The main function is `MultivariateNormalDegenerate._sample_n`, which allows us to sample like this:

```python
import jax.random as jrd
import jax.numpy as jnp
import liesel.model as lsl

key = jrd.PRNGKey(42)
D = jnp.diff(jnp.eye(5))

mvnd = lsl.MultivariateNormalDegenerate.from_penalty(loc=0.0, var=1.0, pen=D.T @ D)
mvnd.sample(seed=key)
```

Some notes:

1. I took the opportunity to also refactor the module a bit. I separated the helper function `_rank_and_log_pdet` into `_rank` and `_log_pdet`. Those new functions now take an array of eigenvalues instead of the precision matrix as input. 
2. The original function `_rank_and_log_pdet` had some special behavior to ensure efficiency, i.e., if you supplied both a `rank` and a `log_pdet` argument, the function essentially did nothing and just returned its inputs. I moved this functionality to the new `MultivariateNormalDegenerate.rank` and `MultivariateNormalDegenerate.log_pdet` properties. They are public, because - why not? 
3. I implemented them using the [`@cached_property`](https://docs.python.org/3/library/functools.html#functools.cached_property) decorator. This decorator computes its value once and then saves it as an attribute on the instance. Consecutive calls are then reduced to lookups.
4. I used this decorator again to provide the $QA^{1/2}$ projector matrix (see #34) in the `_sample_n` function. I think this is a nice way to both keep the code clean and relatively easy to read, while making sure that we do not have unnecessary repeated computations. 
5. I'm not entirely sure about the name for the $QA^{1/2}$ property. For the time being, I simply called it `_projector`. If you have a better idea, I'm happy!

Looking forward to your review @hriebl!

### DIFF
--- a/liesel/distributions/mvn_degen.py
+++ b/liesel/distributions/mvn_degen.py
@@ -72,13 +72,13 @@ class MultivariateNormalDegenerate(tfd.Distribution):
     log_pdet
         The log-pseudo-determinant of the precision matrix. Optional.
     validate_args
-        Python ``bool``, default ``False``. When ``True``, distribution parameters \\
-        are checked for validity despite possibly degrading runtime performance. \\ When
-        ``False``, invalid inputs may silently render incorrect outputs.
+        Python ``bool``, default ``False``. When ``True``, distribution parameters \
+        are checked for validity despite possibly degrading runtime performance. \
+        When ``False``, invalid inputs may silently render incorrect outputs.
     allow_nan_stats
         Python ``bool``, default ``True``. When ``True``, statistics (e.g., mean, \
         mode, variance) use the value ``NaN`` to indicate the result is undefined. \
-        When ``False``, an exception is raised if one or more of the statistic's \\
+        When ``False``, an exception is raised if one or more of the statistic's \
         batch members are undefined.
     name
         Python ``str``, name prefixed to ``Ops`` created by this class.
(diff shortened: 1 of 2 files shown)


hriebl (approved):
After a few discussions on Rocket.Chat, this is ready to be merged. Thank you, @jobrachem! :bouquet:

PR #41, opened 2023-03-03
Add doctest action

jobrachem:
Adds a doctest action to our GitHub CI.

I could also imagine that this can simply be executed as an additional step in the existing `pytest-pull.yml` action. However, if it is its own action, we have a clearly defined distinction between failing doctests and failing unit tests, which I think is nice.


hriebl (approved):
Very cool, thank you!

PR #47, opened 2023-03-08
Improve mvn docs

jobrachem:
A screenshot of the changed docs from local.

I also added public versions of the `._loc` and `._prec` attributes for completeness.

image

### DIFF
--- a/liesel/distributions/mvn_degen.py
+++ b/liesel/distributions/mvn_degen.py
@@ -113,7 +113,8 @@ class MultivariateNormalDegenerate(tfd.Distribution):
     to have the desired covariance structure, and 3) add the desired mean.
 
     The main problem is to find out how we have to transform the standard normal samples
-    in step 2. Say that we have a singular precision matrix :math:`P`. We can view it as
+    in step 2. Say that we have a singular :math:`(m \\times m)`  precision matrix
+    :math:`P`. We can view it as
     the generalized inverse of a variance-covariance matrix :math:`\\Sigma`. We can
     obtain :math:`\\Sigma` by finding the eigenvalue decomposition of the precision
     matrix, i.e.
@@ -141,9 +142,14 @@ class MultivariateNormalDegenerate(tfd.Distribution):
 
                         & = Q A Q^T \\
 
-                        & = \\Sigma
+                        & = \\Sigma.
 
     In the last step, we add the desired mean :math:`\\mu` to :math:`x`.
+    Note that the distribution is not a proper distribution on :math:`\\mathbb{R}^m`,
+    where :math:`m` refers to the number of columns and rows of :math:`P`.
+    Any vector in the null space of :math:`P` can be added to any
+    :math:`x \\in \\mathbb{R}^m` without changing the density. The samples generated
+    using the procedure described above are orthogonal to the null space of :math:`P`.
     """
 
     def __init__(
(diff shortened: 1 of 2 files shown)


jobrachem:
@wiep this is ready for review. No need to go through the sampling source code, please rather review the description in the documentation for correctness and clarity. Thank you!

liesel/distributions/mvn_degen.py
```diff
@@ -95,8 +103,37 @@ class MultivariateNormalDegenerate(tfd.Distribution):
...
+
+    If we have a singular precision matrix :math:`P`, we can view it as the
+    pseudo-inverse :math:`\\Sigma^+` of the variance-covariance matrix :math:`\\Sigma`.
+    We can obtain :math:`\\Sigma` by finding the singular value decomposition
+
+    .. math::
+        \\Sigma^+ = QA^+Q^T,
+
+    where :math:`Q` is the orthogonal matrix of eigenvectors and :math:`A^+` is the
+    diagonal matrix of singular values of :math:`\\Sigma^+ = P`. If the precision matrix
```
wiep, line 119: P is square. Thus, I would refer to EVD instead of SVD. Also, above you refer to eigenvals and not SVs.

liesel/distributions/mvn_degen.py
```diff
@@ -95,8 +103,37 @@ class MultivariateNormalDegenerate(tfd.Distribution):
...
+    is singular, then :math:`\\text{diag}(A^+)` contains zeroes. Now we take the inverse
+    of the non-zero entries of :math:`\\text{diag}(A^+)`, while the zero entries remain
+    at zero, resulting in a matrix :math:`A`. We can now write
+
+    .. math::
+        \\Sigma = Q A Q^T.
+
+    To sample from :math:`N(\\mu, \\Sigma)`, we first draw a sample
```
wiep, line 127: $N(\mu, \Sigma)$ is not a proper normal distribution since Sigma is singular.  Maybe you can call it $N_D$ or $N^\perp$.  It's probably easiest to follow Rue and Helds notation.

liesel/distributions/mvn_degen.py
```diff
@@ -95,8 +103,37 @@ class MultivariateNormalDegenerate(tfd.Distribution):
...
+    :math:`z \\sim N(0, I)` from a standard normal distribution. Next, we transform the
+    sample by applying :math:`x = Q A^{1/2}z`, such that
+    :math:`\\text{Cov}(x) = \\Sigma`:
```
wiep, line 130: Can't you remove the entries corresponding to zero EVs from Q and A and reduce the dimension of z?

wiep (commented):
I think it is overall well comprehensible. I added a few comments where the text could be a bit more precise. Check Algo 3.1 and above in the Rue and Held book.

liesel/distributions/mvn_degen.py
jobrachem: Rue and Held seem to avoid using a specific Notation for a singular normal distribution. Thomas uses the same notation as for a non-singular normal distribution in the appendix of his book, (B.3.2). I might just rephrase the sentence to avoid the symbol, too.

liesel/distributions/mvn_degen.py
jobrachem: Yep, that leads to identical results. We could add that as an issue to further improve the efficiency of the sampling function. For the moment, I think the gain in efficiency would not be so big, so personally, I would prefer to leave the function as it is at the moment - it might be a bit of a haggle with the batch sizes again.

jobrachem:
@wiep Here's the updated version:

image

wiep (approved):

PR #48, opened 2023-03-10
Improve docstrings for Node and Var, closes #43

### LINKED ISSUE #43: Clarify the distinction between the different Node types and Vars in the documentation
I find that I still have some trouble to think about the distinction between a `Node` and a `Var`. I think it would be helpful for me to have some guidance about the distinction in the respective docstrings.

hriebl:
This PR improves the docstrings for the Node and the Var, trying to clarify when it makes sense to declare a Var and when not. It also extends the docstrings of Obs and Param.

@jobrachem, please take a look and let me know if there's anything else you would like to see in the docstrings.

### DIFF
--- a/liesel/model/nodes.py
+++ b/liesel/model/nodes.py
@@ -122,7 +122,39 @@ class NodeState(NamedTuple):
 
 
 class Node(ABC):
-    """A node of a computational graph."""
+    """
+    A node of a computational graph that can cache its value.
+
+    Liesel represents statistical models as directed acyclic graphs (DAGs) of
+    random variables (see :class:`.Var`) and computational nodes. The graph of
+    random variables is built on top of the computational graph. The nodes of the
+    computational graph will typically express computations in JAX returning arrays
+    or pytrees_, but in general, they can represent arbitrary operations in Python.
+
+    Nodes can cache the result of the operations they represent, improving
+    the efficiency of the graph. The cached values are part of the model state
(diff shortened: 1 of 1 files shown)


liesel/model/nodes.py
```diff
@@ -631,7 +663,28 @@ def value(self) -> float:
...
+    e.g. fixed data like hyperparameters or design matrices, or quantities that are
+    computed from other nodes, e.g. structured additive predictors in semi-parametric
+    regression models.
+
+    When a :class:`.Node` does not have an associated probability distribution, it is
+    not strictly necessary to declare it as a variable. There is no hard and fast rule
+    when such a node should be declared as a variable and when not. The advantages of
+    declaring a node as a variable are: (1) easier access via the :attr:`.Model.vars`
+    attribute, and (2) more explicit visualization with the :func:`.plot_vars`
+    function.
```
jobrachem, line 681: I feel like this makes it sound a bit more liberal than it is. The `Node`  child classes like `Dist`, `Data`, `Calc` seem to have rather narrow purposes, and the `Node` cannot be used directly.

jobrachem:
@hriebl the docs are coming along nicely I think. I added some comments with respect to some details.

hriebl:
I've committed the updated docstrings. Please take another look, @jobrachem. :eyes:

jobrachem (approved):

PR #49, opened 2023-03-10
Adds a method to convert sampling results to arviz's inference data

wiep:
Basic implementation is completed. Left to do:

- [x] add an option save the warmup
- [x] documentation
- ~allow to pass additional chains similar to the summary object~

edit: for now, i will not implement the additional chains since this feature is kind of obsolete with next-gen liesel and the upcoming goose update. If the user has an additional chain they could easily add it to the inference data similar to the following code:

<details><summary>code</summary>

```py
import numpy as np
import arviz as az
gen = np.random.default_rng(0)

ps = {
    "x": gen.normal(0, 1, (3, 100, 3, 2)),
    "y": gen.normal(0, 1, (3, 100)),
    "z": gen.normal(0, 1, (3, 100, 2)),
}

additional = {
    "xsq": ps['x']**2,
}

# creates InferenceData object similar to the output of the function from liesel
infdat = az.from_dict(
    posterior=ps
)

# convert the additional chain into an inference data object
additionl_infdat = az.convert_to_inference_data(
    additional
)

# uses the merge function of xarray.DataSet
infdat.posterior = infdat.posterior.merge(additionl_infdat.posterior)

# display the summary of the updated data
az.summary(infdat)
```
</details>


wiep:
@jobrachem, the code in the test file is largely identical to the code in `test_summary_m.py`. If you know how it could be used in both places without being copied please let me know or change it.

jobrachem:
@wiep Thanks, it looks good already! I added the following changes:

- extracted the `MockKernel` to its own module in the tests for easier sharing across test files, 
- moved the shared fixtures to `conftest.py`
- moved `conftest.py` inside the `tests/` directory, such that it can cleanly import code from `tests/mock_kernel.py`
- moved tests for the arviz functionality to `tests/experimental`
- created `__init__.py` files to let the relevant test directories correctly behave as packages

### Log-likelihood group

I have one request: It would be very comfortable if there was a dedicated way to fill the [`"log_likelihood"` group of the InferenceData object](https://python.arviz.org/en/0.14.0/schema/schema.html#log-likelihood) with the pointwise log-likelihood values from sampling - if these are saved. Having these in the InferenceData object lets us easily calculate the WAIC, which would be very useful. You can see my first steps towards this with my pre-liesel-v0.2 CTM code here: https://github.com/liesel-devs/liesel-ctm/issues/46#issuecomment-1405028521. I'm not sure what is the best solution for this usecase, since I think the log-likelihood values will only be available if they are explicitly requested in the `.positions_included` attribute. So I guess three options would be. They are ordered in increasing radicality.

1. Document an intended way of filling the `"log_likelihood"` group
2. Try to fill the group automatically, don't complain if the needed values are not found
3. Make it the default to have the pointwise log-likelihood evaluations available in the `SamplingResults`

wiep:
Currently, we do not have a standardized way to access the log-likelihood, log-prior, or point-wise log-likelihood. Changing that is part of the goose redesign.

I want to avoid putting some code in this function with unique interpretations for specific position keys. Instead, I would leave it to the user to reinterpret the returned inference data. Extracting the point-wise log-likelihood and reinterpreting it as in the log_likelihood group requires only three lines of code. Of course, we can provide a convenience function and deprecate it once the goose redesign has landed. See the example below.

Would you be ok with that?

```py
import arviz as az
import numpy as np
import scipy.stats

obs = np.random.normal(0, 1, (10,))

size = (3, 100, 1)
loc_chain = np.random.normal(0, 0.1, size)
loglik_pointwise = scipy.stats.norm.logpdf(obs, loc=loc_chain, scale = 1)

idat = az.convert_to_inference_data(
    {
        "loc": loc_chain,
        "loglik_pointwise": loglik_pointwise,
    }
)

# idat is what the implemented function should give you

# re-interpret the data
llpw_extracted = idat.posterior['loglik_pointwise']
idat.posterior = idat.posterior.drop('loglik_pointwise')
idat.add_groups({'log_likelihood': {'observed': llpw_extracted}})
az.waic(idat)

# >>> Computed from 300 posterior samples and 10 observations log-likelihood matrix.
# >>>                Estimate    SE
# >>> elpd_waic.    -17.43      3.40 
# >>> p_waic        0.16          -
```

jobrachem:
Sounds reasonable to me. I think putting an example like the one you posted here into the docstring would be enough until the goose redesign 😊

wiep:
@jobrachem, the documentation is extended and covers now this case.

jobrachem (approved):

PR #50, opened 2023-03-10
Add convert_dtype() method to GraphBuilder

hriebl:
This PR adds a convert_dtype() method to the GraphBuilder, which should be pretty self-explanatory. The method is particularly useful when facing dtype-related errors from TFP, typically because NumPy arrays (64-bit most of the time) are used with JAX-TFP (expecting 32-bit). Here's a minimal example:

``` python
import liesel.model as lsl
import numpy as np

a = lsl.Var(0.0, name="a")
b = lsl.Var(np.array(0.0), name="b")
c = lsl.Var("foo", name="c")

gb = lsl.GraphBuilder()
gb.add(a, b, c)

print((a.value, b.value, c.value))
## (0.0, array(0.), 'foo')

gb.convert_dtype("float32")

print((a.value, b.value, c.value))
## (0.0, array(0., dtype=float32), 'foo')
```

If we decide that we want to have such a method, I would continue to work on this PR and add some unit tests.

### DIFF
--- a/liesel/model/model.py
+++ b/liesel/model/model.py
@@ -341,6 +341,32 @@ class GraphBuilder:
 
         return model
 
+    def convert_dtype(self, dtype: str | jax.numpy.dtype) -> GraphBuilder:
+        """
+        Tries to convert the node values in the graph to the specified data type.
+
+        Works for nodes whose value is an array or pytree_. Nodes whose value is of
+        another type are silently ignored.
+
+        .. _pytree: https://jax.readthedocs.io/en/latest/pytrees.html
+        """
+        nodes, _ = self._all_nodes_and_vars()
+
+        def try_astype(x):
+            try:
+                return x.astype(dtype, copy=False)
+            except AttributeError:
+                return x
+
+        for node in nodes:
+            try:
+                value = jax.tree_map(try_astype, node.value)
+                node.value = value  # type: ignore # data node
+            except AttributeError:
+                pass
+
+        return self
+
     def copy(self) -> GraphBuilder:
         """Returns a shallow copy of the graph builder."""
         gb = GraphBuilder()


wiep:
would this method covert an integer (let's say int64) to a float32? If so, I suggest that we function has two arguments `from` and `to`.

hriebl:
It would do any conversion that the astype() methods from NumPy and JAX-NumPy would do, so yes, by default it would convert integers to floats, see [the NumPy docs](https://numpy.org/doc/stable/reference/generated/numpy.ndarray.astype.html).

liesel/model/model.py
```diff
@@ -341,6 +341,37 @@ def build_model(self, copy: bool = False) -> Model:
...
+                    return x
+            except AttributeError:
+                return x
+
+        for node in nodes:
+            try:
+                value = jax.tree_map(try_astype, node.value)
+                node.value = value  # type: ignore # data node
+            except AttributeError:
+                pass
```
jobrachem, line 371: In which cases do we run into an AttributeError here? Only happens, when the node does not have a `.value` attribute, right? When does that happen?

jobrachem (commented):
I like it. Some comments:

- Earlier, we talked about the idea to automatically call this method in `GraphBuilder.build_model()`. I still think this would cover a lot of use cases. But I think it would not cover the parameter transformation use case.  So maybe actually it would be nice to call `try_astype(_from="float64", _to="float32")` on the specific node that is being transformed? That, in addition to calling the `convert_dtype` method in `build_model`, woudl cover 95% of the pain-points, I think.

liesel/model/model.py
hriebl: I happens if the property doesn't have a setter.

jobrachem: I see. Thanks 👍

wiep (commented):
I like it. One minor point: I would rename the arguments. For me anything that starts with an underscore reads "internal". We could use `from_type`, `to_type`. `to` is actually not even a keyword so we could also use `from_` and `to` but then it looks inconsistent. I don't mind having longer argument names since most editors have code completion.

Options:

- `from_type`, `to_type` ⭐ 
- `from_`, `to`
- `from_`, `to_`
- `_from`, `_to`

jobrachem:
I added my preferences for the argument names to Paul's comment.

wiep:
Since @hriebl wrote the code, you can decide ;-)

jobrachem:
- int64 -> int32?
- message about conversion
- add conversion to .add() ?

hriebl:
I've addressed the open comments, so this should be ready for review now.

hriebl:
Not really, other than not deciding between jnp and np, although:
```
>>> jnp.float32 is np.float32
False
>>> jnp.float32 == np.float32
True
```

jobrachem (approved):

jobrachem:
Looks good to me, too. Merging now, since both reviewers approved.

PR #51, opened 2023-03-15
Refactor distregbuilder

jobrachem:
This PR includes only the mildest changes that I have in mind. For the other ones, I think it is more productive to open a separate PR and discuss those changes separately.

### Specifically, the changes are:

1. The mandatory order in which methods have to be called is reversed now (add_response -> add_predictor -> add_smooth). Closes https://github.com/liesel-devs/liesel-internal/issues/212
2. All model-building methods of the distregbuilder now return `self`, such that method-chaining is completely allowed. Closes https://github.com/liesel-devs/liesel-internal/issues/142

In [#212](https://github.com/liesel-devs/liesel-internal/issues/212), we talked about exchanging `add_prediction` for a mandatory argument in `add_response`. When working on the builder though, I felt like this would make usage more clunky than before, because we also have to account for the `inverse_link` of the predictor.

Now, an example for usage of the DistRegBuilder would be:

```python
drb = (
    dr.DistRegBuilder()
    .add_response(y, tfd.Normal)

    .add_predictor("loc", tfb.Identity)
    .add_predictor("scale", tfb.Exp)

    .add_p_smooth(X, m=0.0, s=0.0, predictor="loc", name="xloc")
    .add_p_smooth(X, m=0.0, s=0.0, predictor="scale", name="xscale")
)
```

### DIFF
--- a/liesel/model/distreg.py
+++ b/liesel/model/distreg.py
@@ -4,6 +4,8 @@ Distributional regression.
 
 from __future__ import annotations
 
+from collections import defaultdict
+
 import jax.numpy as jnp
 import jax.random
 import numpy as np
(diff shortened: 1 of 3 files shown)


jobrachem:
Corresponding update to rliesel: https://github.com/liesel-devs/rliesel/pull/1

hriebl (approved):
Thank you, @jobrachem! Merging this after a short discussion with Johannes.

PR #57, opened 2023-03-24
Use () instead of <> in __repr__

wiep:
When running the code below in a jupyter notebook, the name of the variable that raised the exception is not shown. This is due to a display issue in jupyter that hides `<name>`. I suggest to use `(name)` instead.

This PR uses parenthesis instead of angle brackets in the `__repr__` functions in liesel model to fix the issue.

```py
import liesel.model as lsl

p = lsl.Param(1.0, lsl.Dist(tfd.HalfNormal(1.0)), name='p')
n0 = lsl.Var(lsl.Calc(lambda x: 1/0 * x, x=p), name='n0')

gb = lsl.GraphBuilder()
gb.add(n0)
gb.transform(p)
gb.build_model()
```


hriebl:
I used `Var<name>` originally, because `Var(name)` looks like a call that reproduces the variable. How about `Var(..., name='name')` instead?

wiep:
The PR is ready for review. I can change the code to `Var(name="name")` instead of `Var(name="name", ...)` if you think we should have that. Otherwise, there is probably not much to review.

liesel/model/nodes.py
```diff
@@ -1171,4 +1171,4 @@ def __getitem__(self, key) -> Var | Node:
         return self._nodes_and_vars[key]
 
     def __repr__(self) -> str:
-        return f"{type(self).__name__}<{self.name}>"
+        return f"{type(self).__name__}({self.name})"
```
jobrachem, line 1174: I think for the group, it would be nice to have the same style as for the Var, i.e. `Group(name="name")`

liesel/model/model.py
```diff
@@ -98,10 +98,10 @@ class GraphBuilder:
     >>> c = lsl.Var(lsl.Calc(lambda x, y: x + y, a, b), name="c")
     >>> gb = lsl.GraphBuilder()
     >>> gb.add(c)
-    GraphBuilder<0 nodes, 1 vars>
+    GraphBuilder(0 nodes, 1 vars)
     >>> model = gb.build_model()
     >>> model
-    Model<9 nodes, 3 vars>
+    Model(9 nodes, 3 vars)
```
jobrachem, line 104: For the model, this checks out. It's clear what the information in the repr refers to 👍

jobrachem (commented):
@hriebl and I had a brief chat about the repr just now and came out in favor of the variant with `...`. Would you change the variable repr to `Var(name="name")` @wiep ? Thanks! 

I think for consistency's sake it would be nice to also update the repr of the `Node` class accordingly (and for the `Group`, as commented above).

jobrachem (approved):

PR #62, opened 2023-03-30
add tutorials overview page

jobrachem:
Nothing all too fancy, but an overview page. The toc in the sidebar can be expanded.

image

Closes https://github.com/liesel-devs/liesel-internal/issues/244

PR #63, opened 2023-03-30
Add groups tutorial

jobrachem:
Closes https://github.com/liesel-devs/liesel/issues/46

Builds on https://github.com/liesel-devs/liesel/pull/62, so that should be reviewed and merged first.

Here's a preview:

image

### DIFF
--- a/docs/source/tutorials/qmd/07-groups.qmd
+++ b/docs/source/tutorials/qmd/07-groups.qmd
@@ -302,7 +302,7 @@ class VarianceIG(lsl.Group):
         self.tau2 = lsl.Param(start_value, distribution=prior, name=name)
         """Variance parameter variable."""
 
-        super().__init__(name=name, a=a_var, b=b_var, tau2=tau2)
+        super().__init__(name=name, a=self.a, b=self.b, tau2=self.tau2)
 ```
 
 This concludes our tutorial on group usage.
--- a/docs/source/tutorials/qmd/07-groups.qmd
+++ b/docs/source/tutorials/qmd/07-groups.qmd
@@ -1,0 +1,308 @@
+---
+engine: knitr
+---
+
+# Advanced Group Usage
+
+This is a tutorial on how liesel's {class}`.Group` class can be used to derive new
+classes that represent re-usable model parts. This tutorial covers the following parts:
+
+1. Derive new classes from {class}`.Group` to represent related {class}`.Var`s.
+2. Use these groups to create a Gibbs kernel.
+
+
+## Background
+
+As the topic for this group tutorial, we will create groups for a semiparametric regression model
+using Bayesian P-Splines. In this tutorial, we only give minimal background on the model itself - instead, for details on the model, the interested reader should consider the original paper by [Lang & Brezger (2012)](https://doi.org/10.1198/1061860043010).
+
+In Bayesian P-Splines, we are working with a rank-deficient (generate) multivariate
+normal prior on a vector of spline coefficients $\boldsymbol{\beta}$. The unnormalized prior can be written as
+
+$$
+p(\boldsymbol{\beta}) \propto \exp\left(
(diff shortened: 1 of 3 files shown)


jobrachem:
@wiep would you do the review?

docs/source/tutorials/qmd/07-groups.qmd
```diff
@@ -0,0 +1,330 @@
...
+the `tau2_group`. The latter would be created automatically, just as the
+`coef_group` is in the code snippet above.
+
+### Group graph
+
+To see what we have created, let's initialize a toy example of our group to look
+at the resulting graph. First, we create the group:
+
+```{python}
+basis_matrix = np.random.uniform(size=(15, 10)) # NOT a valid basis matrix, just for show!
```
wiep, line 216: We should change this eventually to a real design_matrix.

docs/source/tutorials/qmd/07-groups.qmd
```diff
@@ -0,0 +1,330 @@
...
+1. Here, we create a `SplineCoef` instance *inside* the new group's `__init__`, because
+   this step is always the same with P-Splines, once we know the penalty and the variance
+   parameter $\tau^2$.
+2. We do not automatically create an instance of our `IGVariance` group for the variance
+   parameter, but allow users to pass any fitting group instance to the class. This makes
+   our `PSpline` class very flexible: we could easily replace the `IGVariance` group
+   with another group that represents a different prior for $\tau^2$. For this `PSpline`
+   group, it is only important that there is a key `"tau2"` in the `tau2_group`.
+3. In our `super().__init__` call at the end, we make sure to add not only the
+   Vars `basis_matrix` and `smooth` that we directly created, but also all
```
wiep, line 175: i like the term design matrix more than basis matrix but i can live with it.

i don't like the term smooth. can we use something different? for example f_of_x or effect_x

docs/source/tutorials/qmd/07-groups.qmd
```diff
@@ -0,0 +1,330 @@
...
+new classes that represent re-usable model parts. This tutorial covers the
+following parts:
+
+1. Derive new classes from {class}`.Group` to represent related {class}`.Var`s.
+2. Use these groups to create a Gibbs kernel.
+
+
+## Background
+
+As the topic for this group tutorial, we will create groups for a semiparametric
```
wiep, line 17: i would also briefly discuss that the spline is used to approximate $f(x)$ as $X \beta$ and what the purpose of the prior on $tau$ is. just two sentences. later $X$ and $X \beta$ are used but the reader might not be familiar.

wiep (approved):
Nice work. It looks pretty comprehensible to me. I added a few comments. Please address them as you deem appropriate and then merge the PR.

docs/source/tutorials/qmd/07-groups.qmd
jobrachem: I agree. I intended to use `scipy`s spline functionality, but I did not find a nice helper function for creating the knots - and pasting a function for that into the tutorial seemed a bit much to me. I think if/once we have the spline functionality form liesel-internal in liesel-core, we should certainly replace this here. Then we can also complete the example in the tutorial by fitting a real model.

docs/source/tutorials/qmd/07-groups.qmd
jobrachem: I'll do that.

docs/source/tutorials/qmd/07-groups.qmd
jobrachem: I can see that, because the term "smooth" does not really speak to me, either. I chose it, because it fits with the terminology used throughout the `distreg` module. For the time being, I would tend to stick with `smooth`, but add an explanatory sentence.

Issue #37, opened 2023-02-22
Add helper function to simulate data from a model graph


hriebl:
Here's a first draft of such a function:

``` python
import jax.numpy as jnp
import jax.random
import liesel.model as lsl
import networkx as nx


def _build_simulation_graph(nodes):
    edges = []

    for node in nodes:
        for _input in node.all_input_nodes():
            if isinstance(node, lsl.Dist) and _input is node.at:
                edges.append((node, _input))
            else:
                edges.append((_input, node))

    graph = nx.DiGraph(edges)
    graph.add_nodes_from(nodes)
    return graph


def simulate_model(model, seed, skip=()):
    graph = _build_simulation_graph(model.nodes.values())
    nodes = list(nx.topological_sort(graph))

    dists = [node for node in nodes if isinstance(node, lsl.Dist)]
    seeds = jax.random.split(jax.random.PRNGKey(seed), len(dists))

    for dist, seed in zip(dists, seeds):
        if dist.name in skip:
            continue

        tfp_dist = dist.init_dist()
        event_shape = tfp_dist.event_shape
        batch_shape = tfp_dist.batch_shape
        value_shape = jnp.asarray(dist.at.value).shape
        sample_index = len(value_shape) - len(batch_shape) - len(event_shape)
        sample_shape = value_shape[:sample_index]

        if isinstance(dist.at, lsl.TransientIdentity):
            dist.at.inputs[0].value = tfp_dist.sample(sample_shape, seed)
        else:
            dist.at.value = tfp_dist.sample(sample_shape, seed)
```

``` r
library(rliesel)
```

    Please set your Liesel venv, e.g. with use_liesel_venv()

``` r
set.seed(1337)

n <- 100
x <- runif(n)
y <- rnorm(n, mean = sin(2 * pi * x), sd = 0.2)
plot(x, y)
```

![](https://user-images.githubusercontent.com/5891134/220754287-10ff6080-184c-41cd-8025-60c530681fee.png)

``` r
model <- liesel(
  response = y,
  distribution = "Normal",
  predictors = list(
    loc = predictor(~s(x), inverse_link = "Identity"),
    scale = predictor(~1, inverse_link = "Exp")
  )
)
```

``` python
model = r.model
lsl.plot_vars(model)
```

![](https://user-images.githubusercontent.com/5891134/220754291-a233454f-dc4d-4111-891a-961cb38bd3f9.png)

``` python
# reduce the prior variance of the scale parameter for better simulation results
[...]

PR #70, opened 2023-07-03
Implement Model.simulate()

hriebl:
Implements a slightly improved version of the function in #37 as a method of the model class. Includes docstrings and tests.

### DIFF
--- a/liesel/model/model.py
+++ b/liesel/model/model.py
@@ -14,6 +14,7 @@ from typing import IO, Any, TypeVar
 
 import dill
 import jax
+import jax.numpy as jnp
 import jax.random
 import networkx as nx
 import tensorflow_probability.substrates.jax.bijectors as jb
@@ -34,6 +35,7 @@ from .nodes import (
     NodeState,
     TransientIdentity,
     Var,
+    VarValue,
 )
 from .viz import plot_nodes, plot_vars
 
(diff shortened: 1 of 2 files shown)


liesel/model/model.py
```diff
@@ -955,6 +976,74 @@ def set_seed(self, seed: jax.random.KeyArray) -> Model:
...
+        for dist, seed in zip(dists, seeds):
+            tfp_dist = dist.init_dist()
+
+            event_shape = tfp_dist.event_shape
+            batch_shape = tfp_dist.batch_shape
+            value_shape = jnp.asarray(dist.at.value).shape  # type: ignore
+            sample_index = len(value_shape) - len(batch_shape) - len(event_shape)
+            sample_shape = value_shape[:sample_index]
+
+            value = tfp_dist.sample(sample_shape, seed)
```
jobrachem, line 1032: I think that here, edge cases could occur: It is not required that users implement the `._sample_n` method of a custom TFP class. In that case, they will receive an error:

```
NotImplementedError: sample_n is not implemented
```

I think this behavior is ok. Do you agree @hriebl ?

jobrachem (commented):
Everything looks clear and functional to me, this is some really nice functionality! 👍 I just added a single comment and a little test to convince myself that the function operates as I expected it to with hierarchical models.

I'm wondering a little about the workflow, maybe you can help me to better understand how it is intended. 

Say, I want to simulate lots of data from my prior. What is the intended way to create multiple simulation samples? How would you recommend to postprocess the data for further analysis?

liesel/model/model.py
hriebl: I would agree that this is the expected behavior. The user can then either implement a sampling method or skip the corresponding node with the `skip` argument.

PR #71, opened 2023-07-06
Implement use of monitor flag for liesel.models

wiep:
I've implemented a hacky version to support the monitor flag specified in a liesel model in goose's EngineBuilder. It is not in the final stage because I want to try the implementation before creating a new Protocol class.

~~Currently, the change makes the test `test_sample_transformed_model` in `tests/model/test_goose.py` fail. I assume that is because the tuning requires the key to be tracked `mu_value` is not tracked.~~

### DIFF
--- a/liesel/goose/builder.py
+++ b/liesel/goose/builder.py
@@ -78,6 +78,8 @@ class EngineBuilder:
         """List of position keys that should not be tracked. Excluded keys override
         additional keys."""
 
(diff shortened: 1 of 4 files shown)


jobrachem:
I'd be okay with closing this PR for the time being. While a monitor flag would be convenient, it was not important enough for us in the last two years to move forward. We can always reopen.

PR #72, opened 2023-07-12
Goose-based Initialization strategies

GianmarcoCallegher:


GianmarcoCallegher:
@hriebl I don't know if I should use `Option` for `_initi_startegies`. If so, at some point I should get the value of the object. This can be done using the property `value` of the class. This works for some tests but for others I get the following error: `AttributeError: 'dict' object has no attribute 'value'`

liesel/goose/builder.py
```diff
@@ -57,15 +65,17 @@ class EngineBuilder:
...
         self._engine_key: KeyArray = keys[1]
+        self._init_strategy_key: KeyArray = keys[2]
         self._num_chains: int = num_chains
         self._kernels: list[Kernel] = []
         self._quantity_generators: list[QuantityGenerator] = []
         self._model_state: Option[ModelState] = Option(None)
-
         self._model: Option[ModelInterface] = Option(None)
+        # self._init_strategies: Option[InitStrategy] = Option({})
+        self._init_strategies: InitStrategies = {}
```
hriebl, line 78:
```py
self._init_strategies: Option[InitStrategy] = Option(None)
```

liesel/goose/builder.py
```diff
@@ -124,6 +134,14 @@ def set_initial_values(self, model_state: ModelState, multiple_chains=False):
 
         self._model_state = Option(model_states)
 
+    def set_init_startegies(self, init_strategies: InitStrategies):
+        self._init_strategies = init_strategies
```
hriebl, line 138:
```py
def set_init_strategies(self, init_strategies: InitStrategies | None):
    self._init_strategies = Option(init_strategies)
```

Also note your typo!

liesel/goose/builder.py
```diff
@@ -124,6 +134,14 @@ def set_initial_values(self, model_state: ModelState, multiple_chains=False):
...
+
+    @property
+    def init_startegies(self) -> InitStrategies:
```
hriebl, line 141:
```py
def init_startegies(self) -> Option[InitStrategies]:
```

Probably, we should rather return `self._init_strategies.value` here to avoid user-exposed options, but let's keep it as is for now for consistency with the other properties.

liesel/goose/builder.py
```diff
@@ -222,9 +237,53 @@ def build(self) -> Engine:
...
+            if pos_key not in init_strategies.keys():
+                init_strategies[
+                    pos_key
+                ] = lambda key, current_value: current_value + jax.random.uniform(
+                    key=key,
+                    shape=current_value.shape,
+                    dtype=current_value.dtype,
+                    minval=-1.0,
+                    maxval=1.0,
+                )
```
hriebl, line 266: Don't jitter by default. If `init_strategies.is_none()`, just skip jittering altogether. If  `init_strategies.is_some()`, jitter the parameters that have a strategy, and skip those that don't have one with a warning.

hriebl (commented):
Thank you for your work! I've added a few first comments. One more thought: Maybe we should not call the attribute `self._init_strategies` but `self._jitter_fns`, etc. I know I came up with the original name, but `self._jitter_fns` seems shorter and more descriptive.

liesel/goose/builder.py
GianmarcoCallegher: Should I raise a warning if they are not set?

hriebl: You mean if `init_strategies.is_none()`? Yes, maybe that's a good idea too...

GianmarcoCallegher: Sure, but we should also handle that in the `DistRegBuilder`

GianmarcoCallegher: Ready for a second check

GianmarcoCallegher:
Ready for a second check

GianmarcoCallegher:
Ready for the second check. There are 2 warnings generated by some incompatible type with the new version of `jax`. See #73

hriebl (changes requested):
Thank you for the update! I've fixed a typo and a few minor things, plus there are two more comments below.

liesel/goose/builder.py
```diff
@@ -222,9 +237,51 @@ def build(self) -> Engine:
...
+
+            current_position = model.extract_position(jitter_fns_pos_keys, model_states)
+            jittered_position = {}
+
+            for i, pos_key in enumerate(jitter_fns_pos_keys):
+                jittered_position[pos_key] = jitter_fns[pos_key](
+                    jitter_keys[i], current_position[pos_key]
+                )
+
+            model_states = jax.vmap(model.update_state)(jittered_position, model_states)
```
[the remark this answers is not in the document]
GianmarcoCallegher: I confirm that, with the current implementation, all chains get a different starting value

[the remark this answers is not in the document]
GianmarcoCallegher: This code works and all the chains will get a different value since `current_position[pos_key]` usually has dimension `num_chains` x `param.shape`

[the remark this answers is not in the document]
GianmarcoCallegher: Probably the new implementation is less error-prone

hriebl:
Is this ready for another round of review, @GianmarcoCallegher?

GianmarcoCallegher:
> Is this ready for another round of review, @GianmarcoCallegher?

Yes

hriebl (approved):
Thank you for providing the additional tests, @GianmarcoCallegher! This is now ready to be merged.

PR #74, opened 2023-07-29
close #73

### LINKED ISSUE #73: Make liesel compatible with jax v0.4.14
In Jax 0.4.14, the `DeviceArray` class has been removed (see [changelog](https://github.com/google/jax/blob/jax-v0.4.14/CHANGELOG.md)). As a result, liesel is incompatible with jax 0.4.14.

We use the DeviceArray class for type hints in three places in `goose/summary_viz.py`. These should be changed to `jax.Array`.

jobrachem:
Simply replace the three remaining uses of `jnp.DeviceArray` with `jax.Array` to establish compatibility with Jax v0.4.14. See #73

### DIFF
--- a/liesel/goose/summary_viz.py
+++ b/liesel/goose/summary_viz.py
@@ -6,7 +6,7 @@ from collections.abc import Sequence
 from typing import Any
 
 import arviz
-import jax.numpy as jnp
+import jax
 import matplotlib.pyplot as plt
 import numpy as np
 import pandas as pd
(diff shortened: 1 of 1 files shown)


jobrachem:
@hriebl would you review?

hriebl (approved):
I confirmed that `jax.Array` is available since v0.4.1, which is the current lower bound of our dependency, so we're good to go!

GianmarcoCallegher:
If one installs liesel from pip, they will get the same error. I guess we have to release a new version

PR #75, opened 2023-08-01
Automatic Variable Transformation

GianmarcoCallegher:


GianmarcoCallegher:
Ready for the review, @hriebl

hriebl (changes requested):
Hey Gianmarco, maybe I'm being silly, but do we need that many changes in the model module? Why wouldn't something like the version I implemented in PR #77 be sufficient?

hriebl:
Closing this in favor of #77.

PR #76, opened 2023-08-02
Create a visual distinction for data flow into Calculators and Distributions

GianmarcoCallegher:


jobrachem:
I would like to see what the result looks like ;)

GianmarcoCallegher:
```

GianmarcoCallegher:
> I'm not sure whether I correctly understand what the purpose is. Is this interpretation correct?
> 
> Arrows that represent input to a variable's distribution are colored in light grey now, to create a visual distinction from arrows that represent input to a variable's calculator. The latter are colored in black, like they used to be.

Yes

jobrachem:
Thanks! Maybe then a legend entry for both cases would be nice, something like "black: Used in value", "grey: Used in distribution". What do you think?

GianmarcoCallegher:
> Thanks! Maybe then a legend entry for both cases would be nice, something like "black: Used in value", "grey: Used in distribution". What do you think?

Sure

GianmarcoCallegher:
You can check now

jobrachem:
image

I like it. One additional question: Why do you write "Used in value or calculator"? Can an input node be used in another node's value without the use of a calculator?

GianmarcoCallegher:
> image
> 
> I like it. One additional question: Why do you write "Used in value or calculator"? Can an input node be used in another node's value without the use of a calculator?

You are right, it's redundant. I fixed it

PR #77, opened 2023-08-10
Automatic variable transformation with less changes

hriebl:
Adaptation of PR #75 requiring less changes in the models module.


GianmarcoCallegher:
Everything looks good and cleaner than my version, I just simplified `make_transformed_distribution`

hriebl:
@GianmarcoCallegher, I think 192ea292b76ee22bf60344bc8dc3c5b5ffd3ae31 is not equivalent to the old version: If the default event space bijector depends on the arguments of the TFP distribution (which is unlikely but not impossible), the old version used the bijector given the current model state, while the new version uses the bijector given the model state from when `transform()` was called...

Issue #78, opened 2023-08-14
Improve error messages for Node.update()


jobrachem:
I often find it hard to identify which node causes a problem, because when there's an error in a node's update() method, I have no information about which node it is. This is an example error message:

```python
>>> gb.plot_vars()
Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "/Users/johannesbrachem/Documents/git/liesel/liesel/model/model.py", line 490, in plot_vars
    model = Model(nodes_and_vars, grow=False)
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/johannesbrachem/Documents/git/liesel/liesel/model/model.py", line 796, in __init__
    node.update()
  File "/Users/johannesbrachem/Documents/git/liesel/liesel/model/nodes.py", line 509, in update
    self._value = self.function(*args, **kwargs)
                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "<string>", line 1, in <lambda>
ValueError: matmul: Input operand 1 has a mismatch in its core dimension 0, with gufunc signature (n?,k),(k,m?)->(n?,m?) (size 9 is different from 8)
```

I think in general, it would be helpful to include at least the node's name in the error message. Should be quite straight-forward to implement.

PR #84, opened 2023-09-05
Address #78

jobrachem:
If an error occurs in a call to `Calc.update()`, the exception is now re-raised with a printout of the instance repr.

See #78


jobrachem:
@hriebl I just requested you as the reviewer mostly because Paul is busy and Gianmarco on vacation - I hope that's fine for you.

wiep:
Should this error get it's own type instead of the most general `RuntimeError`?

I guess the question is rather if we should start to define error types for liesel in general.

jobrachem:
Another option that would not require us to define new error types could be to raise the same error as the original one and just append something to its message.

PR #85, opened 2023-09-05
Correct sampling for sigma_sq

jobrachem:
Addresses https://github.com/liesel-devs/liesel-internal/issues/282
This is how it looks:

image


hriebl (approved):
Without rereading the full tutorial, the changes look good to me!

PR #100, opened 2023-09-27
Adapt to blackjax 1.0.0

wiep:
Adapts the code for the hmc and nuts kernels to reflect changes in blackjax 1.0.0.

BlackJax changed their kernel interface. They use now classes with the member function `step`. I adapted the code such that it requires minimal changes in our code base. However, the readability is decreased and we should tidy it up either in this PR or in a subsequent PR.

### DIFF
--- a/liesel/goose/hmc.py
+++ b/liesel/goose/hmc.py
@@ -4,7 +4,6 @@ Hamiltonian/Hybrid Monte Carlo (HMC).
 
 from collections.abc import Callable, Sequence
 from dataclasses import dataclass, field
-from functools import partial
 from typing import ClassVar
 
 import jax.numpy as jnp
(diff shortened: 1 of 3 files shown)


jobrachem:
Awesome, that was so quick! Are you still actively working on this or should I already take a look tomorrow? If so, you can remove the draft status.

hriebl (approved):
Thanks, @wiep! I've tidied up your code a little bit and think this is now ready to be merged.

jobrachem (approved):
After @hriebl's added tidying, I now also looked at the changes and additionally ran the tests locally in a new venv. I think this is ready to be merged.

wiep:
thanks :-) that was fast 🚀

PR #102, opened 2023-10-17
Fix Mypy errors for untyped functions

GianmarcoCallegher:


GianmarcoCallegher:
I'm not sure if this is the best way of fixing it, but it works.

jobrachem (approved):
Looks good! Thanks! Also nice that you added the option  `disable_error_code = "annotation-unchecked"`

Issue #129, opened 2023-11-01
Remove deprecations for v0.4.0


jobrachem:
- [x] liesel.model.legacy.Param
- [x] liesel.model.legacy.Obs
- [x] liesel.goose.engine.SamplingResult
- [x] liesel.goose.summary_m.Summary.from_result
- [x] liesel.goose.summary_m.Summary.from_results
- [x] `liesel.model.goose.GooseModel` -> (renamed/moved) `liesel.goose.interface.LieselInterface`
- [x] `liesel.goose.models.DictModel` -> (renamed/moved) `liesel.goose.interface.DictInterface`
- [x] `liesel.goose.models.DataClassModel` -> (renamed/moved) `liesel.goose.interface.DataclassInterface`

wiep:
Is this list up-to-date?

jobrachem:
- [ ] `liesel.goose.engine.stack_for_multi` deprecated since 0.1.0
- [ ] `GraphBuilder.transform` deprecated in 0.2.9
- [ ] `lsl.obs`, `lsl.param` deprecated in 0.4.0
- [ ] `lsl.add_group` deprecated in 0.2.2

PR #130, opened 2023-11-01
Rename Param and Obs

### LINKED ISSUE #53: Rename helper functions: Param -> param
We have some helper functions that are capitalized at the moment. Users might expect them to be classes because of Python's naming conventions. This issue is about renaming them to avoid this confusion.

- [ ] Param -> param
- [ ] Obs -> obs
- [ ] Anything else?

jobrachem:
Closes #53 

Notes:

- Getting the documentation right was a little tricky, because Sphinx does not work well with two objects of the same name (except for case differences) in the same module.
- To deal with this, I moved `Param` and `Obs` to the `liesel.model.legacy` module. 
- The import paths for users are not affected. I made sure that `Param` and `Obs` are imported to `liesel.model.__init__.py`, so they can still be used as `lsl.Param` and `lsl.Obs`
- The two functions are marked as deprecated, issue `FutureWarnings` and are marked to be removed in v0.4.0, see #129


jobrachem:
Hi @wiep, would you do the review on this one?

wiep (approved):
looks good to me. do you want me to build the documentation locally?

jobrachem:
I think that's not necessary. 

I just added a changelog entry. If you approve, feel free to merge and delete the branch.

wiep:
great. Just a note: We need to publish a new version when we update the manuscript on arxiv so that the manuscript and software are in sync.

PR #133, opened 2023-11-02
MVN Degenerate from smoothing param

GianmarcoCallegher:


GianmarcoCallegher:
@jobrachem would you review?

liesel/distributions/mvn_degen.py
```diff
@@ -206,7 +206,7 @@ def __init__(
         )
 
     @classmethod
-    def from_penalty(
+    def from_penalty_variance(
```
jobrachem, line 209: I would strongly prefer a solution that only adds functionality and does not change the existing API.

jobrachem (changes requested):
See comment.

jobrachem (approved):
Thanks Giammy! :)

liesel/distributions/mvn_degen.py
jobrachem: I see what you mean - still, thanks for reverting back to the old name 😊

PR #135, opened 2023-11-02
Fix group tutorial (closes #131)

### LINKED ISSUE #131: Group usage tutorial: The rank seems to be overlooked in the model graph
_Originally posted by @Seb-Lorek in https://github.com/liesel-devs/liesel/discussions/105#discussioncomment-7443219_

Actually as far as I can see, I need to add the group. Otherwise the rank of the smooth is not added to the `model.state` dictionary, which I need for the `tau2`Gibbs Kernel. Here the code (excluding the group set up [see here](https://docs.liesel-project.org/en/latest/tutorials/md/07-groups.html) for details)

```
# Set up model in liesel
# Fixed parameter prior
beta_loc = lsl.Var(0.0, name="beta_loc")
beta_scale = lsl.Var(100.0, name="beta_scale")

# Set up the fixed parameters
beta_dist = lsl.Dist(tfjd.Normal, loc=beta_loc, scale=beta_scale)
beta = lsl.Param(value=np.array([0.0]), distribution=beta_dist, name="beta")

# Set up the smooth parameters
tau2_group = VarianceIG(name="tau2", a=1.0, b=0.0005)

penalty = X.smooth_pen_mat_cent[0]
smooth_group_1 = PSpline(name="smooth_1", basis_matrix=X.design_mat_cent[1], penalty=penalty, tau2_group=tau2_group)

# Set up the scale 
sigma_a = lsl.Var(0.01, name="sigma_a")
sigma_b = lsl.Var(0.01, name="sigma_b")

sigma_dist = lsl.Dist(tfjd.InverseGamma, concentration=sigma_a, scale=sigma_b)
sigma = lsl.Param(value=10.0, distribution=sigma_dist, name="sigma")

Z = lsl.Obs(X.fixed_data, name="Z")

lpred_loc_fn = lambda z, beta, smooth_1: jnp.dot(z, beta) + smooth_1
lpred_loc_calc = lsl.Calc(lpred_loc_fn, z=Z, beta=beta, smooth_1=smooth_group_1["smooth"])

lpred_loc = lsl.Var(lpred_loc_calc, name="lpred_loc")

response_dist = lsl.Dist(tfjd.Normal, loc=lpred_loc, scale=sigma)
response = lsl.Var(y.to_numpy(), distribution=response_dist, name="response")
```

Building the model without `gb.add_groups(smooth_group_1)`

```
gb = lsl.GraphBuilder().add(response)
gb.transform(sigma, tfjb.Exp)

model = gb.build_model()
model
```

Setting up the `EngineBuilder` ( for the setup of the Gibbs Kernel `tau2_gibbs_kernel` for `tau2` [see here](https://docs.liesel-project.org/en/latest/tutorials/md/07-groups.html#use-group-to-build-a-gibbs-sampler))

```
builder = gs.EngineBuilder(seed=1337, num_chains=4)

builder.set_model(lsl.GooseModel(model))
builder.set_initial_values(model.state)

builder.add_kernel(tau2_gibbs_kernel(smooth_group_1))
builder.add_kernel(gs.IWLSKernel(["smooth_1_coef"]))
builder.add_kernel(gs.IWLSKernel(["beta"]))
builder.add_kernel(gs.NUTSKernel(["sigma_transformed"]))

builder.set_duration(warmup_duration=1000, posterior_duration=1000)

builder.positions_included = ["sigma"]

engine = builder.build()
```

Finally calling 

```
engine.sample_all_epochs()
```

results in an error due to not finding the rank of the smooth coefficients in the `model_state` object in the `EngineBuilder`.

jobrachem:
There was a subtle coding error that led to unexpected behavior. See #131 . Now it's fixed.

### DIFF
--- a/docs/source/tutorials/qmd/07-groups.qmd
+++ b/docs/source/tutorials/qmd/07-groups.qmd
@@ -113,8 +113,6 @@ class SplineCoef(lsl.Group):
     def __init__(self, name: str, penalty: Array, tau2: lsl.param) -> None:
         penalty_var = lsl.Var(penalty, name=f"{name}_penalty")
 
-        # we save the rank of the penalty matrix here to use it later in our
-        # gibbs kernel
         evals = jax.numpy.linalg.eigvalsh(penalty)
         rank = lsl.Data(jnp.sum(evals > 0.0), _name=f"{name}_rank")
         _log_pdet = jnp.log(jnp.where(evals > 0.0, evals, 1.0)).sum()
--- a/docs/source/tutorials/qmd/07-groups.qmd
+++ b/docs/source/tutorials/qmd/07-groups.qmd
@@ -333,7 +333,7 @@ class VarianceIG(lsl.Group):
         self.b = lsl.Var(b, name="f{name}_b")
         """Prior scale variable."""
 
-        prior = lsl.Dist(tfd.InverseGamma, concentration=a_var, scale=b_var)
+        prior = lsl.Dist(tfd.InverseGamma, concentration=self.a, scale=self.b)
         self.tau2 = lsl.param(start_value, distribution=prior, name=name)
         """Variance parameter variable."""
 
--- a/docs/source/tutorials/qmd/07-groups.qmd
+++ b/docs/source/tutorials/qmd/07-groups.qmd
@@ -110,18 +110,23 @@ Next, we create the group for our spline coefficient in a very similar way.
 
 ```{python}
 class SplineCoef(lsl.Group):
-    def __init__(self, name: str, penalty: Array, tau2: lsl.Param) -> None:
+    def __init__(self, name: str, penalty: Array, tau2: lsl.param) -> None:
         penalty_var = lsl.Var(penalty, name=f"{name}_penalty")
 
         # we save the rank of the penalty matrix here to use it later in our
         # gibbs kernel
-        rank = lsl.Data(np.linalg.matrix_rank(penalty), _name=f"{name}_rank")
+        evals = jax.numpy.linalg.eigvalsh(penalty)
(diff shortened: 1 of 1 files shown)


jobrachem:
@GianmarcoCallegher same question here: Would you review?

GianmarcoCallegher:
> @GianmarcoCallegher same question here: Would you review?

Same answer

docs/source/tutorials/qmd/07-groups.qmd
```diff
@@ -109,22 +110,25 @@ Next, we create the group for our spline coefficient in a very similar way.
...
+        _log_pdet = jnp.log(jnp.where(evals > 0.0, evals, 1.0)).sum()
+        log_pdet = lsl.Data(_log_pdet, _name=f"{name}_log_pdet")
 
         prior = lsl.Dist(
             MultivariateNormalDegenerate.from_penalty,
             loc=0.0,
             var=tau2,
             pen=penalty_var,
+            rank=rank,
+            log_pdet=log_pdet
```
GianmarcoCallegher, line 127: Maybe we want to make the functions to compute the rank an the log pseudo determinant public?

jobrachem: Fair point. I thought about that too, but the problem that I see is: from a code-organization point of view, I think it could be confusing and hard to predict to export a `rank` and `log_pdet` function from a module that provides a distribution class. And probably we do not really want to get in the business of providing specific linear algebra functions publicly at the moment. So my approach currently would be to not touch that.

GianmarcoCallegher: Very true. I mean, we also have the utility functions for the IWLS, maybe, in the future we could have a separate module for this kind of functions

GianmarcoCallegher (approved):

PR #136, opened 2023-11-02
Refactor goose-models to goose-interfaces (closes #110)

### LINKED ISSUE #110: Rename lsl.GooseModel to lsl.GooseInterface
I think the current naming is confusing and can be improved with the proposed change.

This change would of course include backwards-compatibility: The current lsl.GooseModel would remain functional, it would just be equipped with a deprecation warning before it is fully removed.

jobrachem:
Substantially, this PR contains the following changes:

- `liesel.model.goose.GooseModel` -> (renamed/moved) `liesel.goose.interface.LieselInterface`
- `liesel.goose.models.DictModel` -> (renamed/moved) `liesel.goose.interface.DictInterface`
- `liesel.goose.models.DataClassModel` -> (renamed/moved) `liesel.goose.interface.DataclassInterface`

Everything else are just changes that take this change into account.

Further notes:

- The old names are still available, but issue a `FutureWarning` and are scheduled for removal in v0.4.0 (see #129). 
- The deprecation is also reflected in the documentation


jobrachem:
@wiep can you do the review?

liesel/goose/models.py
wiep, line 1: `goose.models` is not private so the module should not be removed without a deprecation period. I would keep the old models.py to reexport the interfaces using the old names until version 0.4.0.

wiep (changes requested):
looks good to me. i just want to keep the models module until 0.4.0.

liesel/goose/models.py
jobrachem: I can see that. 

I would argue that the API that users can expect to be stable is given by the objects exported from `liesel.goose` or `import liesel.goose as gs`, and `gs.DictModel` etc. is provided.

But I can see your point, and don't see harm in still providing `goose.models`.

jobrachem: I just pushed a commit that restores the goose.models module.

liesel/goose/models.py
```diff
@@ -1,63 +1,37 @@
...
-    A model interface for a model state represented by a ``dict[str, Array]`` and a
-    corresponding log-probability function.
+    Alias for :class:`.DictInterface`, provided for backwards compatibility.
+
+    .. deprecated:: v0.2.6
+        Use :class:`.DictInterface` instead. This alias will be removed in v0.4.0.
     """
 
     def __init__(self, log_prob_fn: LogProbFunction):
         self._log_prob_fn = log_prob_fn
```
wiep, line 15: We should here just call the init method of super and not reimplement it. Similar for the others.

If you think that makes no sense ignore my comment.

wiep (approved):
Thanks. I think it's good. you can consider the comment but also ignore it.

liesel/goose/models.py
jobrachem: I implemented it, I think it's a sensible suggestion. Thanks!

PR #147, opened 2023-11-07
Fix some pytets warnings

GianmarcoCallegher:


GianmarcoCallegher:
The `UserWarning` can be fixed in several ways ()[look here for more details](https://stackoverflow.com/questions/56656777/userwarning-matplotlib-is-currently-using-agg-which-is-a-non-gui-backend-so?page=1&tab=scoredesc#tab-top) by they all involve installing a `pip` package or `tkinter`. I would not do that since they appear only in the tests. Maybe we can just ignore the warnings in the tests. What do you think, @jobrachem ?

tests/model/test_goose.py
jobrachem, line 1: This test file tests the deprecated `lsl.GooseModel`. Changing this code to use `gs.LieselInterface` defeats the purpose a little :D The tests should remain targeted at `lsl.GooseModel` and be removed in #129.  Maybe here we should also just filter the FutureWarnings.

jobrachem (changes requested):
Makes sense! Just one issue, see the comment on `test_goose.py`

tests/model/test_goose.py
GianmarcoCallegher: Sure, that would be better

PR #151, opened 2023-11-10
Implementes NamedTupleInterface

### LINKED ISSUE #150: Implement the NamedTupleInterface
I want to use NamedTuples in the workshop but we do not have a `NamedTupleInterface`. The implementation is pretty straightforward and could be included in Liesel's next release. It needs to be documented and tested.

```python
class NamedTupleInterface:
    
    def __init__(self, log_prob_fn: Callable[[NamedTuple], float]):
        self._log_prob_fn = log_prob_fn

    def extract_position(
        self, position_keys: Sequence[str], model_state: NamedTuple
    ):
        return {key: getattr(model_state, key) for key in position_keys}
    
    def log_prob(self, model_state: NamedTuple) -> float:
        return self._log_prob_fn(model_state)

    def update_state(self, position, model_state: NamedTuple):
        new_state = model_state._replace(**position)
        return new_state
```

wiep:
Closes #150.

Documentation is missing but should be very similar to DictInterface or DataclassInterface.

### DIFF
--- a/liesel/goose/interface.py
+++ b/liesel/goose/interface.py
@@ -337,3 +337,18 @@ class LieselInterface:
             A dictionary of node names and their corresponding :class:`.NodeState`.
         """
         return model_state["_model_log_prob"].value
+
+
+class NamedTupleInterface:
+    def __init__(self, log_prob_fn: LogProbFunction):
+        self._log_prob_fn = log_prob_fn
+
+    def extract_position(self, position_keys: Sequence[str], model_state: ModelState):
+        return {key: getattr(model_state, key) for key in position_keys}
+
+    def log_prob(self, model_state: ModelState) -> float:
+        return self._log_prob_fn(model_state)
+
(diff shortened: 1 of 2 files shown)


wiep:
please squash and merge after review.

liesel/goose/interface.py
jobrachem, line 1: The docstrings for `update_state` and `log_prob` are swapped.

jobrachem (changes requested):
Looks nice! The only thing is that the docstrings for two methods are swapped, see the specific comment. 

I think it also makes sense to list the new interface in `docs/source/goose.rst`, between lines 38 and 39, to include it here: https://docs.liesel-project.org/en/stable/goose.html#set-up-your-model

liesel/goose/interface.py
wiep: 🙈

wiep:
comments addressed. please squash and merge.

PR #156, opened 2023-11-15
Allow passing key as seed

Seb-Lorek:
Simple proposal to allow `seed` argument to be a key generated from `jax.random.PRNGKey` in the initialization of class `EngineBuilder`. Unfortunately type testing for the key is not really strict with `jax.Array`. However using `jnp.issubdtype(seed.dtype, jnp.uint32)` makes it more complicated. Stricter type testing seems only possible with `jax.random.key` currently, see [#16716](https://github.com/google/jax/issues/16716) and [#16781](https://github.com/google/jax/pull/16781) for current development on this.

### DIFF
--- a/liesel/goose/builder.py
+++ b/liesel/goose/builder.py
@@ -130,7 +130,7 @@ class EngineBuilder:
     from your posterior distribution.
     """
 
-    def __init__(self, seed: int | jax.Array, num_chains: int):
+    def __init__(self, seed: int | KeyArray, num_chains: int):
         if isinstance(seed, int):
             keys = jax.random.split(jax.random.PRNGKey(seed), 3)
         elif isinstance(seed, jax.Array):
--- a/liesel/goose/builder.py
+++ b/liesel/goose/builder.py
@@ -72,6 +72,7 @@ class EngineBuilder:
     Parameters
     ----------
     seed
+        Either an int or a key generated from jax.random.PRNGKey.
         Used for jittering initial values and MCMC sampling.
     num_chains
         The number of chains to be used.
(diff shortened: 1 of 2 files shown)


liesel/goose/builder.py
```diff
@@ -129,8 +130,16 @@ class EngineBuilder:
     from your posterior distribution.
     """
 
-    def __init__(self, seed: int, num_chains: int):
-        keys = jax.random.split(jax.random.PRNGKey(seed), 3)
+    def __init__(self, seed: int | jax.Array, num_chains: int):
```
wiep, line 133: the type annotation for a key should be `jax.random.KeyArray`. currently, that is just a type alias to array but it might change in the future.

liesel/goose/builder.py
```diff
@@ -129,8 +130,16 @@ class EngineBuilder:
...
+        if isinstance(seed, int):
+            keys = jax.random.split(jax.random.PRNGKey(seed), 3)
+        elif isinstance(seed, jax.Array):
```
wiep, line 136: please test for `KeyArray`.

wiep (changes requested):
Thanks for submitting a PR. I made two comments in line about the type of key.

It would be great if you could add a test such that we know both cases work. I would expect, that the builders have the same state if they are initialized with a int or a key generated from that seed. Below, `builder` and `builder2` should have the same state.

```python
int_seed = 0
key_seed = jax.random.PRNGKey(int_seed)
builder = EngineBuilder(seed=int_seed, ...)
builder2 = EngineBuilder(seed=key_seed, ...) 
```

So I would suggest to write add whitebox test, where you make sure that the fields 
`_prng_key`, `_engine_key`, and `_jitter_key` have the same value. The function jnp.all can be used to test if all values in a bool array are true.

liesel/goose/builder.py
Seb-Lorek: I used this at first however I got a `DeprecationWarning`, that's why I added the two issues above that deal with the switch away from using `jax.random.KeyArray` or `jax.random.PRNGKeyArray`. Here again some infos from the [docs](https://jax.readthedocs.io/en/latest/jep/9263-typed-keys.html#type-annotations-for-prng-keys).

liesel/goose/builder.py
Seb-Lorek: See above.

liesel/goose/builder.py
wiep: I missed that jax has implemented a new key style. I thought they just added more PRNG generators. To be consistent with the other code in Goose, i think it would be best to use `from liesel.goose.typing import KeyArray` which is just an alias to `typing.Any` and then do the test you proposed.

@jobrachem, at some point in the future, we need to change that everywhere in liesel. Also, we should use jax new jax new type Annotation. In Goose, we apparently use a type alias to Any as Array.

https://github.com/liesel-devs/liesel/blob/1ee02de035199d52448eb8788b13a720db5aafb6/liesel/goose/types.py#L22C8-L28

wiep (approved):
can be merged if checks pass

jobrachem:
I just merged the PR. Thanks @Seb-Lorek and congrats on your first code contribution to the liesel repo! 🎉  😊

PR #166, opened 2023-12-01
fix name `basis_matrix` in 07-groups.qmd

Seb-Lorek:
Add format string to the name of `basis_matrix`, to allow for the unique definition of several B-spline basis matrices.


jobrachem (approved):
Looks good, thank you @Seb-Lorek!

PR #167, opened 2023-12-04
Raise error if duplicate nodes are detected in GraphBuilder.transform

GianmarcoCallegher:


jobrachem:
Hey @GianmarcoCallegher, thanks for your PR! 😊 I have two suggestions:

1. A unit test that confirms the expected behavior would be nice
2. It would be nice if you would directly add a bullet point to the `CHANGELOG.md` about the change.

liesel/model/model.py
```diff
@@ -214,7 +214,13 @@ def _all_nodes_and_vars(self) -> tuple[list[Node], list[Var]]:
...
         """
         nodes = self.nodes.copy()
-        nodes.extend(node for var in self.vars for node in var.nodes)
+        # nodes.extend(node for var in self.vars for node in var.nodes)
+
+        for var in self.vars:
+            for node in var.nodes:
+                if node not in nodes:
+                    nodes.append(node)
+
```
jobrachem, line 223: What does this change accomplish? Isn't it pretty much the same as before? Why is line 217 commented out instead of deleted?

jobrachem (changes requested):
See comments

liesel/model/model.py
GianmarcoCallegher: Right. Its is useless. I removed it

GianmarcoCallegher:
> Hey @GianmarcoCallegher, thanks for your PR! 😊 I have two suggestions:
> 
> 1. A unit test that confirms the expected behavior would be nice
> 2. It would be nice if you would directly add a bullet point to the `CHANGELOG.md` about the change.

Done it

jobrachem (approved):

PR #170, opened 2023-12-07
Rename Data -> Value; keep old name as an alias

jobrachem:
This PR goes back to a suggestion made by @Seb-Lorek :

>I find the naming of the node "Data" a bit unfortunate. I would have thought of Fixed or something like that. When you think of Data, you generally think of oberved data in the model. Also the description: "A Node subclass that holds constant data.". Doesn't help, it reminded me more of observed data. (Thought the description somehow fit: "Constant values better.")

I am not sure whether I am entirely happy with `Const`, because this class is also used as the `lsl.Var.value_node` of strong variables. If they are parameters, the values are not actually constant, but updated during sampling. So maybe `Value` would be better than `Const`. What do you guys think @GianmarcoCallegher @wiep? Of course this PR is not super urgent, so no need to prioritise it right now,

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -180,3 +180,4 @@ cython_debug/
 
 # MacOS
 .DS_Store
+/tests/notebooks
--- a/docs/source/model.rst
+++ b/docs/source/model.rst
@@ -35,7 +35,7 @@ sure to check them out.
     :nosignatures:
 
     ~liesel.model.nodes.Var
-    ~liesel.model.nodes.Data
+    ~liesel.model.nodes.Const
     ~liesel.model.nodes.Calc
     ~liesel.model.nodes.Dist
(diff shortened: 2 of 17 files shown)


wiep:
The name `Const` would suggest that the value is constant and immutable. From the two options, I like `Value` better.

GianmarcoCallegher:
Is this ready for a review?

jobrachem:
No, the name change is not implemented in the way it is supposed to be.

jobrachem:
@GianmarcoCallegher this is finally ready for review 😊

jobrachem:
Since the docs actions had not completed for a while, I pushed a minor commit (an addition to the gitignore) to trigger a new build.

Issue #89, opened 2023-09-06
Line color in model graph is wrong when a variable is transformed


jobrachem:
Take the example from this tutorial: https://docs.liesel-project.org/en/latest/tutorials/md/01a-transform.html

The graph after transformation looks like this:

![](https://docs.liesel-project.org/en/latest/_images/graph-and-transformation-1.png)

- The black edges pointing from `a` and `b` to `sigma` are correct, because the model accounts for the possibility that these parameters *could* be used in the bijector.
- The black edges pointing from `a` and `b` to `sigma_transformed` are wrong. They should be grey, because these parameters are parameters of the distribution.

It is currently not clear whether this is an error in the plotting functionality or in the underlying model functionality.

hriebl:
Maybe this problem is best solved by splitting `Var.all_input_vars()` into `Var.all_value_input_vars()` and `Var.all_dist_input_vars()`. Then [this check](https://github.com/liesel-devs/liesel/blob/main/liesel/model/viz.py#L168-L170) can be reduced to `if edge[0] in edge[1].all_dist_input_vars(): ...` etc. Otherwise too much complex logic is needed in the plotting functions to find out where a `Var` really points...

wiep:
@jobrachem , the plot above seems to be find now. I think have fixed the issue.

PR #174, opened 2023-12-27
Make transform a method

### LINKED ISSUE #93: Refactor GraphBuilder.transform() method
Notes from a discussion with @hriebl:

Currently, the `GraphBuilder.transform()` method covers two cases at the same time:

1. Users explicitly supply a bijector that should be used for the transformation.
2. Users do not explicitly supply a bijector; instead, the default bijector is obtained from the node's tensorflow distribution class.

The idea of this issue is to internally separate the functionality of the transform method into these two parts. The method's API would stay the same.

In the process, there is one nice thing we can achieve with this separation: For bijectors explicitly supplied by users, we can determine whether the model graph still requires edges going from the inputs of the original node's distribution to the original node, or whether it is enough to have the edges point to the transformed node's distribution. Consider the model graph below. In this graph, we could omit the edges from `a` and `b` to `sigma`, because `a` and `b` are only used in `sigma_transformed` after the transformation.

When we are automatically obtaining the default bijector based on the node's tensorflow distribution, there does not seem to be any way for us to know whether the mentioned edges are needed in the bijector or not. So, because they *may* be needed, and, on a more technical side, because we always need these inputs to initialize the distribution class to obtain the bijector in the first place, we cannot remove the edges in the second case.

Technical background:

In the internal helper function `_transform_back`, the input parameters of the original node's distribution are fed as inputs in the back-transformation calculator:

    return Calc(fn, var_transformed.value_node, *inputs, **kwinputs)  # type: ignore

https://github.com/liesel-devs/liesel/blob/e3eec6e885173a4d4c7430d58e74abbd8ec4543b/liesel/model/model.py#L81C1-L81C1

### LINKED ISSUE #173: Add transformation method to lsl.Var
I think it would make sense to add a method `lsl.Var.transform(bijector)`. This would transform the var in question and return the new, transformed var.

jobrachem:
This PR introduces a `Var.transform` method. Some notes:

### Replacing `GraphBuilder.transform`

We currently have `GraphBuilder.transform`. The new method is in fact intended as a replacement for `GraphBuilder.transform`. Having `Var.transform` as a method on `Var` has the advantage that it is easier to find for users. Since the transformation is "doing something to a Var" and it does not actually require any functionality within the GraphBuilder, living as a method on a Var is a natural development for the transform method.

### Behavior change

The method behaves similar to `GraphBuilder.transform`, with a few notable differences:

1. It will *not* use the default event space bijector from tensorflow by default. This change is made to encourage users to either select their desired bijector manually, which is often sensible, or to request an automatic bijector manually. In both cases, users are more aware of what they are doing.
2. The new method accepts bijector *instances* in addition to bijector *classes*. In fact, passing an instance is the preferred way of passing a bijector. Passing a bijector class is only supported if you actually defined `*bijector_args` or `*bijector_kwargs` to be passed to the bijector. This simplifies the code for the default case. More importantly, this fixes the graph representation after transformation, see below.

Using a bijector instance:

image

Compare to using the default event space bijector. Note that in this case, the same bijector is being used, but there are spurious edges from the nodes "v0" and "v1", the prior parameters, to the original variable "tau".

image

### Deprecation and updated documentation

1. I marked `GraphBuilder.transform` as deprecated and included directions towards the new method in its documentation.
2. I updated the usage of the transformation method in the tutorials `01a-transform.md` and `04-mcycle.md`. As it turned out, usage in these tutorials was outdated anyway.

### Notebook for testing

You can play around with the method in this notebook:

[050-transform.ipynb.zip](https://github.com/liesel-devs/liesel/files/13779942/050-transform.ipynb.zip)

### Related issues

- This PR closes #173 
- This PR closes #93
- This PR improves upon the behavior described in #89, because the problem described in #89 does not occur if a bijector instance is passed.

### DIFF
--- a/liesel/model/model.py
+++ b/liesel/model/model.py
@@ -6,6 +6,7 @@ from __future__ import annotations
 
 import logging
 import re
+import warnings
 from collections import Counter
 from collections.abc import Iterable
 from copy import deepcopy
@@ -110,7 +111,7 @@ class GraphBuilder:
         GraphBuilder.
     :meth:`.GraphBuilder.build_model` : Method for building a model from the
         GraphBuilder.
-    :meth:`GraphBuilder.transform` : Transforms a variable by adding a new transformed
+    :meth:`.Var.transform` : Transforms a variable by adding a new transformed
         variable as an input. This is useful for variables that are constrained to a
         certain domain, e.g. positive values.
 
@@ -295,7 +296,7 @@ class GraphBuilder:
         --------
         :meth:`.GraphBuilder.build_model` : Method for building a model from the \
             GraphBuilder.
-        :meth:`GraphBuilder.transform` : Transforms a variable by adding a new
+        :meth:`.Var.transform` : Transforms a variable by adding a new
             transformed variable as an input.
 
         Examples
@@ -698,6 +699,10 @@ class GraphBuilder:
         """
         Transforms a variable by adding a new transformed variable as an input.
 
+        .. deprecated:: 0.2.9
+            ``GraphBuilder.transform`` is deprecated.
+            Use :meth:`.Var.transform` instead. This method will be removed in v0.4.0.
+
         Creates a new variable on the unconstrained space ``R**n`` with the appropriate
         transformed distribution, turning the original variable into a weak variable
         without an associated distribution. The transformation is performed using
@@ -763,6 +768,12 @@ class GraphBuilder:
         >>> scale.update().log_prob
         0.0
         """
+
+        warnings.warn(
+            "GraphBuilder.transform is deprecated. Use Var.transform instead. "
+            "This method will be removed in v0.4.0",
+            FutureWarning,
+        )
 
         if var.weak:
             raise RuntimeError(f"{repr(var)} is weak")
--- a/liesel/model/model.py
+++ b/liesel/model/model.py
@@ -444,7 +444,13 @@ class GraphBuilder:
 
         for var in _vars:
             if var.auto_transform:
-                gb.transform(var)
+                tname = f"{var.name}_transformed"
+                if tname in nodes or tname in _vars:
+                    raise RuntimeError(
+                        f"Auto-transform of {var} failed, because a variable of the "
+                        f"name {tname} is already present in {gb}."
+                    )
+                var.transform(bijector=None)
 
         gb._set_missing_names()
         gb._add_model_log_lik_node()
--- a/liesel/model/model.py
+++ b/liesel/model/model.py
@@ -37,6 +37,7 @@ from .nodes import (
     TransientIdentity,
     Var,
     VarValue,
+    is_bijector_class,
 )
 from .viz import plot_nodes, plot_vars
 
@@ -785,7 +786,7 @@ class GraphBuilder:
         if var.dist_node is None:
             raise RuntimeError(f"{repr(var)} has no distribution")
 
-        if bijector is not None and not isinstance(bijector, type) and (args or kwargs):
+        if not is_bijector_class(bijector) and (args or kwargs):
             raise RuntimeError(
                 "You passed a bijector instance and "
                 "nonempty bijector arguments. You should either initialise your "
--- a/liesel/model/model.py
+++ b/liesel/model/model.py
@@ -694,7 +694,11 @@ class GraphBuilder:
         return self
 
     def transform(
-        self, var: Var, bijector: type[Bijector] | None = None, *args, **kwargs
+        self,
+        var: Var,
+        bijector: type[Bijector] | Bijector | None = None,
+        *args,
+        **kwargs,
     ) -> Var:
         """
         Transforms a variable by adding a new transformed variable as an input.
@@ -781,6 +785,14 @@ class GraphBuilder:
         if var.dist_node is None:
             raise RuntimeError(f"{repr(var)} has no distribution")
 
+        if bijector is not None and not isinstance(bijector, type) and (args or kwargs):
+            raise RuntimeError(
+                "You passed a bijector instance and "
+                "nonempty bijector arguments. You should either initialise your "
+                "bijector directly with the arguments, or pass a bijector class "
+                "instead."
+            )
+
         # avoid name clashes
         self._set_missing_names()
 
@@ -816,6 +828,17 @@ class GraphBuilder:
                 "and no bijector was given"
             )
 
+        using_bijector_instance = not use_default_bijector and not isinstance(
+            bijector, type
+        )
+
+        if isinstance(bijector, type) and not args and not kwargs:
+            warnings.warn(
+                "You passed a bijector class instead of an instance, but did not "
+                "provide any arguments. You should either provide arguments "
+                "or pass an instance of the bijector class.",
+            )
+
         if isinstance(tfp_dist, jd.Distribution):
             tfd = jd
             tfb = jb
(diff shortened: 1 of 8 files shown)


jobrachem:
A log message on the default bijector being used would be nice.

jobrachem:
@jobrachem will transfer the core benefits of the `Var.transform` method to the `GraphBuilder.transform`: 1) accepting a bijector instance, 2) logging the default event space bijector

jobrachem:
@jobrachem will continue as follows:

1. Implement Var.transform
2. Make API between Var.transform and GraphBuilder.transform consistent
3. Add a note to the docs of GraphBuilder.transform, referring to Var.transform

jobrachem:
I now harmonized the API between `Var.transform` and `GraphBuilder.transform`. Namely:

1. `Var.transform` now has the same default behavior as `GraphBuilder.transform`, i.e. it tries to use the default event space bijector from tensorflow probability.
2. `GraphBuilder.transform` now also accepts a bijector instance.

There is one small difference in how the methods handle it, if the user passes a bijector *class*: `Var.transform` will raise a RuntimeError, if you pass a bijector class, but do not actually use any arguments for the bijector. `GraphBuilder.transform` will emit a `UserWarning` instead, in order to avoid breaking existing code.

jobrachem:
@wiep gentle reminder :)

liesel/model/model.py
```diff
@@ -799,6 +828,17 @@ def transform(
                 "and no bijector was given"
             )
 
+        using_bijector_instance = not use_default_bijector and not isinstance(
+            bijector, type
+        )
+
+        if isinstance(bijector, type) and not args and not kwargs:
```
wiep, line 835: might use

`if isinstance(bijector, type) and not (args or kwargs):`

for better readability.

liesel/model/model.py
```diff
@@ -758,12 +773,26 @@ def transform(
...
+            FutureWarning,
+        )
+
         if var.weak:
             raise RuntimeError(f"{repr(var)} is weak")
 
         if var.dist_node is None:
             raise RuntimeError(f"{repr(var)} has no distribution")
 
+        if bijector is not None and not isinstance(bijector, type) and (args or kwargs):
```
wiep, line 788: can we test here for `isintance(bijector, BijectorType)` instead of `bijector is not None and not isinstance(bijector, type)`?

would make it better comprehensible

liesel/model/nodes.py
```diff
@@ -1100,6 +1103,159 @@ def all_input_vars(self) -> tuple[Var, ...]:
...
+        if isinstance(bijector, type) and not bijector_args and not bijector_kwargs:
+            raise ValueError(
+                "You passed a bijector class instead of an instance, but did not "
+                "provide any arguments for the bijector. You should either provide "
+                "arguments or pass an instance of the bijector class instead."
+            )
+
+        # avoid infinite recursion
+        self.auto_transform = False
+
```
wiep, line 1232: could handle the `bijector is None` case above line 1233 and remove duplicated code:
```
if bijector is None:
            dist_inst = self.dist_node.init_dist()
            bijector = dist_inst.experimental_default_event_space_bijector
            
if isinstance(bijector, type):
            tvar = _transform_var_with_bijector_class(
                self, bijector, *bijector_args, **bijector_kwargs
            )
else: #bijector is bijector instance
   ...
```

liesel/model/nodes.py
```diff
@@ -1100,6 +1103,159 @@ def all_input_vars(self) -> tuple[Var, ...]:
...
+            tvar = _transform_var_with_bijector_class(
+                self, bijector, *bijector_args, **bijector_kwargs
+            )
+        elif bijector is None:
+            dist_inst = self.dist_node.init_dist()
+            bijector = dist_inst.experimental_default_event_space_bijector
+            tvar = _transform_var_with_bijector_class(
+                self, bijector, *bijector_args, **bijector_kwargs
+            )
+        else:
```
wiep, line 1243: Should we test for the correct type by using:

```
elif isinstance(bjector, BijectorType):
    ...
else:
    raise ValueError("bijector must be None | instance of bijectorType | BijectorType")
```

liesel/model/nodes.py
```diff
@@ -1100,6 +1103,159 @@ def all_input_vars(self) -> tuple[Var, ...]:
...
+        >>> scale.update().log_prob
+        0.0
+        """
+        if self.weak:
+            raise RuntimeError(f"{repr(self)} is weak")
+
+        if self.dist_node is None:  # type: ignore
+            raise RuntimeError(f"{repr(self)} has no distribution")
+
+        if isinstance(bijector, type) and not bijector_args and not bijector_kwargs:
```
wiep, line 1223: similar to above
`and not (bijector_args or bijector_kwargs)`

liesel/model/nodes.py
```diff
@@ -1386,6 +1542,92 @@ def __repr__(self) -> str:
...
+        bijector_inv.forward(var.value),
+        transformed_dist,
+        name=f"{var.name}_transformed",
+    )
+
+    var.value_node = Calc(bijector_inst.forward, transformed_var)
+    return transformed_var
+
+
+def _transform_var_with_bijector_class(
```
wiep, line 1577: to confirm my understanding: we cannot just instanciate  the bijector here and then use `_transform_var_with_bijector_instance` because 

1) the inputs may not be available at this time
2) the bijector may change when the inputs to the bijector change

wiep (commented):
Thanks for the work.

I made a few suggestions to make the if statement better comprehensible. I think I would handle the bijector is None case such that I would cast it into a bijector is instance of BijectorType case. Than you need to handle only two cases. See annotations below.

However, both is a bit of a style question. Should you prefer the current version. I'm also ok with that.


Also, please feel free to revert the commit changing doc strings if you do not agree (dfb5217e381b58878c0c99fc901bba0a765ea0a1).

liesel/model/model.py
jobrachem: I don't know about a `BijectorType` object, but I clarified the check by introducing

```python
def is_bijector_class(obj):
    return isinstance(obj, type) and issubclass(obj, jb.Bijector)
```

and using

```python
if not is_bijector_class(bijector) and (args or kwargs):
```

now :)

liesel/model/nodes.py
jobrachem: Yes, and I think point two is the critical one: Even of the input values were available at the time that `.transform` is called, we do not want to use their concrete value here. Their values may change, and as a result, the bijector's behavior may change. So when a bijector class is used, we have to set up a wrapping system that initializes the bijector anew during every update with the current input values.

jobrachem:
Hey @wiep, thanks a lot for your suggestions! I implemented them as closely as possible. Your change to the docstring is fine with me, too 👍 

Your comment about removing the code duplication also alerted me to two issues:

1. `Var.transform` lacked a test for whether the default event space bijector exists. I added that test.
2. There was a subtle error in my implementation of using the default event space bijector. In the case of using the default bijector, we want to account for the possibility that the default bijector can change depending on the distribution's arguments. So the default bijector has to be obtained and initialized anew for every `.update()` call. This was not the case in my previous implementation. Not it is, so this problem is fixed.

liesel/model/nodes.py
jobrachem: I implemented that as follows:

```python
else:
    raise TypeError(
        f"Argument {bijector=} is of invalid type {type(bijector)}."
    )
```

liesel/model/nodes.py
jobrachem: See my comment below, this  suggestion alerted me to a problem in my implementation. Thanks!

PR #175, opened 2023-12-27
A function for applying stochastic gradient descent to a Liesel model

jobrachem:
This PR introduces the function `gs.optim`.

This function provides a quick-and-easy way to apply stochastic gradient descent to a `lsl.Model` - which can be nice for providing a "warm start" for MCMC sampling. All loops and batching are implemented in JAX.

Optimization can be applied on the whole dataset or in batches. By default, the Adam optimizer is used with a fixed learning rate of `0.1`. Users can supply an optimizer of their choice from the `optax` library.

The PR includes documentation and unit tests.

Here's a notebook for playing around:
[051-optim.ipynb.zip](https://github.com/liesel-devs/liesel/files/13780492/051-optim.ipynb.zip)

### DIFF
--- a/liesel/goose/optim.py
+++ b/liesel/goose/optim.py
@@ -256,8 +256,8 @@ def optim_flat(
     Now, we are ready to run the optimization.
 
     >>> result = gs.optim_flat(model, params=["coef"])
-    >>> result.position
-    {'coef': Array([0.52289313, 1.2874348 ], dtype=float32)}
+    >>> {name: jnp.round(value, 2) for name, value in result.position.items()}
+    {'coef': Array([0.52, 1.29], dtype=float32)}
 
     We can now, for example, use ``result.model_state`` in
     :meth:`.EngineBuilder.set_initial_values` to implement a "warm start" of MCMC
--- a/liesel/goose/optim.py
+++ b/liesel/goose/optim.py
@@ -1,51 +1,80 @@
 from collections.abc import Sequence
 from dataclasses import dataclass
+from typing import Any
 
 import jax
 import jax.numpy as jnp
 import numpy as np
 import optax
+import pandas as pd
 
-import liesel.goose as gs
-import liesel.model as lsl
+from ..model import Model, Node, Var
+from .interface import LieselInterface
+from .types import Array, KeyArray, ModelState, Position
 
-from .types import Array, ModelState, Position
+
+def array_to_dict(
+    x: Array, names_prefix: str = "x", prefix_1d: bool = False
+) -> dict[str, Array]:
+    """Turns a 2d-array into a dict."""
+
+    if isinstance(x, float) or x.ndim == 1:
+        if prefix_1d:
+            return {f"{names_prefix}0": x}
+        else:
+            return {names_prefix: x}
+    elif x.ndim == 2:
+        return {f"{names_prefix}{i}": x[:, i] for i in range(x.shape[-1])}
+    else:
+        raise ValueError(f"x should have ndim <= 2, but it has x.ndim={x.ndim}")
 
 
 @dataclass
 class OptimResult:
-    """Holds the results of model optimization with :func:`.optim`."""
+    """Holds the results of model optimization with :func:`.optim_flat`."""
 
-    position: Position
-    """Position dictionary of optimized parameters with their final values."""
     model_state: ModelState
     """Final model state after optimization."""
+    position: Position
+    """Position dictionary of optimized parameters with their final values."""
     iteration: int
     """Iteration counter of the last iteration."""
-    abs_change: float | Array
-    """Absolute change in negative log probability in the last iteration."""
-    rel_change: float | Array
-    """Relative change in negative log probability in the last iteration."""
-    neg_log_prob: float
-    """Negative log probability after the last iteration."""
+    iteration_best: int
+    """Iteration counter of the iteration with lowest loss."""
+    history: dict[str, dict[str, Array] | Array]
+    """History of loss evaluations and, if applicable, intermediate position values."""
+    max_iter: int
+    """Maximum number of iterations."""
+    n_train: int
+    """Number of training observations."""
+    n_test: int
+    """Number of test observations."""
 
 
-def _find_observed(graph: lsl.Model) -> dict[str, lsl.Var | lsl.Node]:
+def _find_observed(model: Model) -> dict[str, Var | Node]:
     obs = {
         var_.name: jnp.array(var_.value)
-        for var_ in graph.vars.values()
+        for var_ in model.vars.values()
         if var_.observed
     }
     return obs
 
 
 def batched_nodes(nodes: dict[str, Array], batch_indices: Array) -> dict[str, Array]:
-    """Returns a subset of the graph state using the given batch indices."""
+    """Returns a subset of the model state using the given batch indices."""
     return jax.tree_util.tree_map(lambda x: x[batch_indices, ...], nodes)
 
 
-def _find_sample_size(graph: lsl.Model) -> int:
-    obs = {var_.name: var_ for var_ in graph.vars.values() if var_.observed}
+def _generate_batch_indices(key: KeyArray, n: int, batch_size: int) -> Array:
+    n_full_batches = n // batch_size
+    shuffled_indices = jax.random.permutation(key, n)
+    shuffled_indices_subset = shuffled_indices[0 : n_full_batches * batch_size]
+    list_of_batch_indices = jnp.array_split(shuffled_indices_subset, n_full_batches)
+    return jnp.asarray(list_of_batch_indices)
+
+
+def _find_sample_size(model: Model) -> int:
+    obs = {var_.name: var_ for var_ in model.vars.values() if var_.observed}
     n_set = {int(np.array(var_.value.shape)[0, ...]) for var_ in obs.values()}
     if len(n_set) > 1:
         raise ValueError(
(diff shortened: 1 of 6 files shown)


setup.cfg
```diff
@@ -31,6 +31,7 @@ install_requires =
     seaborn>=0.13
     tensorflow-probability>=0.17
     tqdm>=4.62
+    optax
```
wiep, line 34: I think it needs a version requirement like the other dependencies.

jobrachem:
We should think about the batching and how/whether it generalizes to different models. right now it assumes iid data.

jobrachem:
paul: maybe give it a different name that clearly identifies that this assumes iid data

jobrachem:
I had a quick look at the code again. This is the function that does the batching. It is applied to `Var` objects that are marked as `Var.observed`. To be honest, I don't see yet how this implies an assumption of iid data. 

```python
def batched_nodes(nodes: dict[str, Array], batch_indices: Array) -> dict[str, Array]:
     """Returns a subset of the graph state using the given batch indices."""
     return jax.tree_util.tree_map(lambda x: x[batch_indices, ...], nodes)
```

wiep:
It assumes that each node marked as observed can be indexed in the same way. 

Consider a hierarchical model (and do not worry about the validity of the estimation):

Let's assume that $Y_{ij}$ is the outcome variable for the $i$-th level-1 unit in the $j$-th level-2 unit. The covariates at level-1 and level-2 are represented by $X_{ij}$ and $W_{j}$, respectively. The model can be represented as follows:

Level-1 model (within groups):
$$Y_{ij} = \beta_{0j} + \beta_{1j}X_{ij} + \epsilon_{ij}$$
where $\epsilon_{ij}$ is the level-1 error term.

Level-2 model (between groups):
$$\beta_{0j} = \gamma_{00} + \gamma_{01}W_{j} + u_{0j}$$
$$\beta_{1j} = \gamma_{10} + \gamma_{11}W_{j} + u_{1j}$$
where $u_{0j}$ and $u_{1j}$ are level-2 error terms.

The design matrices $\boldsymbol{W}$ and $\boldsymbol{X}$, arising from stacking the covariates, have different number of rows and cannot be indexed in the same way.

jobrachem:
Thanks for the clarification @wiep !

> It assumes that each node marked as observed can be indexed in the same way.

Indeed it does. I would call this a "flat" way of indexing, as opposed to "nested" indexing. Pointing this out in a way that cannot be missed or is at least hard to miss would be important. This limitation only applies to batching though, the rest of the function does not make such an assumption. So I am not certain whether changing the name of the function to something like `optim_flat` would be necessary - it seems a bit overly restrictive to me.

Maybe the function can be designed in a way that allows the use of different, also user-supplied batching strategies. I would have to take a closer look at the code again to see whether this may even be easy to implement.


> Consider a hierarchical model (and do not worry about the validity of the estimation):
> 
> Let's assume that Yij is the outcome variable for the i-th level-1 unit in the j-th level-2 unit. The covariates at level-1 and level-2 are represented by Xij and Wj, respectively. The model can be represented as follows:
> 
> Level-1 model (within groups): Yij=β0j+β1jXij+ϵij where ϵij is the level-1 error term.
> 
> Level-2 model (between groups): β0j=γ00+γ01Wj+u0j β1j=γ10+γ11Wj+u1j where u0j and u1j are level-2 error terms.
> 
> The design matrices W and X, arising from stacking the covariates, have different number of rows and cannot be indexed in the same way.

I do see the point that this model, like it is, would not fit into the batching scheme. You can, however, write down the model in a "flat" way that allows the batching to work. Still, of course that does not eliminate the fact that the `batch_size` argument of the proposed `optim` function has this limitation. This makes it even clearer how important it would be to alert users to the way batching is done.

GianmarcoCallegher:
I think that having different batching strategies would be really nice. Consider also the case in which you have a binary response and a really unbalanced dataset with a lot of 0s and only a few 1s. You might want a batching strategy that guarantees to always have some 1s in each batch

jobrachem:
We will use `optim_flat` for now, this also gives us room to favor a different implementation under the name `optim` later without backwards compatibility issues.

jobrachem:
@wiep if you got a notification for this, you can ignore it. @GianmarcoCallegher is doing the review.

liesel/goose/optim.py
```diff
@@ -0,0 +1,233 @@
...
+
+    if batch_size is None:
+        batch_size = n
+
+    n_batches_inside = n // batch_size
+    batch_indices = jnp.arange(0, batch_size)
+    last_batch_size = n % batch_size
+    last_batch_start = n_batches_inside * batch_size
+    last_batch_end = last_batch_start + last_batch_size
+    last_batch_indices = jnp.arange(last_batch_start, last_batch_end)
```
GianmarcoCallegher, line 180: I think it would be better to change the batch indexes at every iteration. This usually helps to avoid local minima and improves generalization. Also, you can have the last batch of the same size as the others by resampling some data points without replacement.

```python
def _generate_batches_idx(slef, key: JaxArray, N: int, batch_size: int) -> JaxArray:
        key_permutation, key_choice = random.split(key, 2)
        batches_idx = jnp.hstack([
            jax.random.permutation(key_permutation, N),
            jax.random.choice(key_choice, N, (batch_size - N % batch_size, ))
        ])
        return jnp.array_split(batches_idx, len(batches_idx) // batch_size)
```

liesel/goose/optim.py
```diff
@@ -0,0 +1,233 @@
...
+        pos = optax.apply_updates(pos, updates)
+        return pos, opt_state
+
+    def _update_cond(val):
+        i, _, _, nlp_old, nlp_new = val
+
+        abs_change = jnp.abs(nlp_old - nlp_new)
+        rel_change = jnp.abs(abs_change / nlp_old)
+
+        return (abs_change > atol) & (rel_change > rtol) & (i < maxiter)
```
GianmarcoCallegher, line 204: Here we are checking the absolute change and the relative change between epoch $t - 1$ and $t$. I think it would be better to introduce a patience of more epochs. It might be that in some cases we are not monotonically decreasing

GianmarcoCallegher:
Do we want to always have the negative log probability as the function to optimize or do we want to provide another parameter which could be the function to optimize, (e.g. RMSE)?

liesel/goose/optim.py
jobrachem: Oh, so this would mean you don't split up the data into batches of a certain size, but you always draw a random sample of the batch size with replacement? So the same data points can get used twice?

liesel/goose/optim.py
jobrachem: That's a very good point. So far, I did not implement patience, because the algorithm runs without creating a history. But probably it should do that, and then it would be possible to implement some patience.

liesel/goose/optim.py
GianmarcoCallegher: You split the data in batches of a certain size but you change the elements in each batch every iteration. It might be that N % batch_size != 0. In that case, you can resample some data points to fill the batch. It is like doing a "data augmentation".

Example:

You have 5 data points (x_1, x_2, x_3, x_4, X_5) and a batch size of 2:

Iteration 1: (x_1, x_3), (x_4, x_2), (x_5, x_3) sampled randomly
Iteration 2: (x_4, x_5), (x_2, x_1), (x_3, x_4) sampled randomly
Iteration3: ...

jobrachem: That sounds good! Thanks a lot! 😊

jobrachem:
> Do we want to always have the negative log probability as the function to optimize or do we want to provide another parameter which could be the function to optimize, (e.g. RMSE)?

Also a really  good point. I guess it would be relatively easy to let users provide their own function for optimization.

liesel/goose/optim.py
jobrachem: @wiep suggests dropping the last, non-full batch instead of filling it via resampling. This is probably more common.

GianmarcoCallegher (changes requested):
You are a good boy. Do the changes ❤️

liesel/goose/optim.py
GianmarcoCallegher: I haven't found a paper where they prove that adding some resample observations to the last batch leads to an unbiased estimator, but here are two example where they do that:

- [neon](https://neon.nervanasys.com/docs/latest/faq.html)
- [CNN Sentence](https://github.com/yoonkim/CNN_sentence/blob/master/conv_net_sentence.py) from this popular paper [Convolutional Neural Networks for Sentence Classification](https://arxiv.org/pdf/1408.5882.pdf%EF%BC%89%E3%80%82%E4%B8%8B%E9%9D%A2%E7%9C%8B%E4%B8%8B%E8%AE%BA%E6%96%87%E4%B8%AD%E7%9A%84%E6%88%AA%E5%9B%BE%EF%BC%9A)

jobrachem: Thanks for the research! :)

jobrachem:
Okay, I finally implemented a big update. Now we have the following:

1. Batching is implemented as suggested by @GianmarcoCallegher (i.e. randomly assembled batches in each iteration) with the modification suggested by @wiep (i.e. if the number of observations is not divisible by the batch size, the last batch is simply dropped in that iteration).
2. The function now saves a history of training and (if applicable) testing loss, as well as a history of parameter values. The latter can be turned off to save memory.
3. Early stopping is now implement with patience.
4. If the users wishes, they can supply a `test_model`. If supplied, the test loss is used to determine early stopping. For fully Bayesian optimization I think using a test model usually be a little weird, because it affects the balance of likelihood and prior. But there might be cases in which one would want to fit a likelihood model with fully constant priors.

Here's a notebook with a minimal example for testing:

[075-optim.ipynb.zip](https://github.com/liesel-devs/liesel/files/14733605/075-optim.ipynb.zip)

Do you think the current defaults are good?
1. Especially the defaults for the `stopper` are quite arbitrarily chosen right now. 
2. `save_position_history=True`
3. `restore_best_position=True`
4. `prune_history=True`

GianmarcoCallegher:
Thank you very much, @jobrachem. 

Some general remarks:
- I would replace the word "test" with the word "validation"
- Usually, early stopping is performed only if there is a validation dataset which is different from the training dataset, but here, if we don't have the validation dataset, we set it to be equal to the train one `model_test = model_test if model_test is not None else model`. If there is no validation set, you just run the model for all epochs on the training set

What do you think? ❤️

jobrachem:
Thank you very much for the speedy review @GianmarcoCallegher ! Your suggestions make sense to me, I will implement them.

GianmarcoCallegher:
> Okay, I finally implemented a big update. Now we have the following:
> 
> 1. Batching is implemented as suggested by @GianmarcoCallegher (i.e. randomly assembled batches in each iteration) with the modification suggested by @wiep (i.e. if the number of observations is not divisible by the batch size, the last batch is simply dropped in that iteration).
> 2. The function now saves a history of training and (if applicable) testing loss, as well as a history of parameter values. The latter can be turned off to save memory.
> 3. Early stopping is now implement with patience.
> 4. If the users wishes, they can supply a `test_model`. If supplied, the test loss is used to determine early stopping. For fully Bayesian optimization I think using a test model usually be a little weird, because it affects the balance of likelihood and prior. But there might be cases in which one would want to fit a likelihood model with fully constant priors.
> 
> Here's a notebook with a minimal example for testing:
> 
> [075-optim.ipynb.zip](https://github.com/liesel-devs/liesel/files/14733605/075-optim.ipynb.zip)
> 
> Do you think the current defaults are good?
> 
> 1. Especially the defaults for the `stopper` are quite arbitrarily chosen right now.
> 2. `save_position_history=True`
> 3. `restore_best_position=True`
> 4. `prune_history=True`

Concerning the default values, I think they are totally fine. Maybe I would not have a default number of iterations, which has to be specified by the user

jobrachem:
Over lunch, @wiep mentioned the practice of rescaling the likelihood when using mini-batching in order to ensure that the prior does not get more power than warranted. I think this should be implemented in our function here, too. It can be used both in general for minibatching, and also in order to correct for a potentially smaller sample size in a validation model.

liesel/goose/optim.py
```diff
@@ -277,13 +297,17 @@ def optim_flat(
     batch_seed = (
         batch_seed if batch_seed is not None else np.random.randint(low=1, high=1000)
     )
-    model_test = model_test if model_test is not None else model
+
+    user_patience = stopper.patience
+    if model_validation is None:
+        model_validation = model_validation if model_validation is not None else model
```
GianmarcoCallegher, line 303: If `model_validation` is None, you can already assign `model_validation = model` without the inline if statement, right?

GianmarcoCallegher (changes requested):
Thanks a lot, @jobrachem . Everything looks great. I left a couple of more comments

liesel/goose/optim.py
jobrachem: Good point! 😅

liesel/goose/optim.py
```diff
@@ -106,34 +106,54 @@ class Stopper:
     max_iter: int
     patience: int
     atol: float = 1e-3
-    rtol: float = 1e-6
+    rtol: float = 1e-12
```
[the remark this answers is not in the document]
jobrachem: It is very strict - with this default, early stopping basically only depends on the absolute tolerance. That is intentional here for the default.

jobrachem:
@GianmarcoCallegher I implemented my changes. If you are satisfied, please squash & merge 😊

liesel/goose/optim.py
```diff
@@ -0,0 +1,551 @@
...
+    )
+    _validate_log_prob_decomposition(
+        interface_validation, position=position, state=model_validation.state
+    )
+
+    # ---------------------------------------------------------------------------------
+    # Define loss function(s)
+
+    likelihood_scalar = n_train / batch_size
+    likelihood_scalar_validation = n_train / n_validation
```
GianmarcoCallegher, line 356: I'm just curious. Why is that?

GianmarcoCallegher:
Everything LGTM, @jobrachem . I just left one final comment (probably it is not an error, I am just confused)

liesel/goose/optim.py
jobrachem: Thanks for the question @GianmarcoCallegher !
Ah the idea here is to rescale the likelihood to keep its ratio to the prior constant, as discussed here: https://github.com/liesel-devs/liesel/pull/175#issuecomment-2030409019

Otherwise, if you use batching or a small test dataset, the prior gets relatively more important than if you were to use the likelihood. 

Does that make sense in your opinion?

jobrachem:
Thanks @GianmarcoCallegher ! See my reply above :)

PR #177, opened 2024-01-31
Use black preview in pre-commit

jobrachem:
This PR simply adds the line `args: [--preview]` to the black configuration for our pre-commit. I think this is nice predominantly because this includes automatic formatting of long strings, which can be quite annoying. But there's another nice stuff, too, see here: https://black.readthedocs.io/en/stable/the_black_code_style/future_style.html For example, dictionary and list formatting was also improved.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -23,6 +23,7 @@ repos:
     rev: 23.10.1
     hooks:
       - id: black
+        args: [--preview]
   - repo: https://github.com/pycqa/flake8
     rev: 6.1.0
     hooks:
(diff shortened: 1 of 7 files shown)


jobrachem:
P.S.: The changes to other files than the pre-commit config are due to black's updated behavior, as far as I can see.

jobrachem:
@GianmarcoCallegher thanks! You can usually safely delete the branches after merging :)

GianmarcoCallegher:
> @GianmarcoCallegher thanks! You can usually safely delete the branches after merging :)

I always forget that. Sorry :(

PR #186, opened 2024-04-16
Turn off update_on_init for special model nodes

jobrachem:
The adoption of #92 caused warnings to be logged by default during model initialization, because the special nodes for the log likelihood and log probability cannot always be updated directly on their initialization. This PR just sets `update_on_init=False` for these nodes, which solves the issue.


GianmarcoCallegher:
Makes sense

PR #187, opened 2024-04-16
Update logging setup

jobrachem:
Small update to make the liesel default logger behave more like you would expect.

- Increased the level of the `"liesel"` logger from `DEBUG` to `INFO`
- Removed the level `INFO` from the default handler for the liesel logger.

Now, you can simply update the logging level like this:

```python
import logging

logger = logging.getLogger("liesel")
logger.setLevel(logging.DEBUG)
```

Before, our setup required that you changed the level on the *handler*, which is less obvious:

```python
import logging

logger = logging.getLogger("liesel")
logger.handlers[0].setLevel(logging.DEBUG)
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@ All notable changes to this project will be documented in this file.
 
 ## [0.2.10] - [unreleased]
 
+- :construction: Update logging setup (#187, @jobrachem)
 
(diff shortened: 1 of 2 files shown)


jobrachem:
@GianmarcoCallegher would you do the review?

GianmarcoCallegher:
Looks good ❤️

PR #190, opened 2024-06-12
Remove filter warnings from pyproject

GianmarcoCallegher:


jobrachem:
@GianmarcoCallegher wow, it is really satisfying to see that now, there are zero warnings in the tests. Thanks!

jobrachem (approved):

PR #191, opened 2024-06-12
Fix tutorials

jobrachem:
The tutorials are now included in rendered form as markdown files, which should make them more robust. The former `docs` action is now an action that simply executes the tutorials but does not process the results any further. This should alert us to errors in the tutorials.


jobrachem:
Uh nice, with the updated docs setup, the docs are automatically built for pull requests, too 😊

jobrachem:
@GianmarcoCallegher would you give it a quick look by a second pair of eyes and then merge?

jobrachem:
I had to disable the tutorials workflow for the time being. The problem is that, on the GitHub action runner, we get reticulate 1.2x, but we need 1.31 or higher for `rliesel`. I would say that is a secondary issue and should not prevent this PR from being merged. I'm opening a new issue.

PR #198, opened 2024-07-24
Unpin scipy

### LINKED ISSUE #185: Compatibility to scipy 1.13
Scipy 1.13 is currently incompatible to Liesel due to changed import paths.

- [x] Update Liesel to be compatible with scipy 1.13
- [x] Unpin dependency

jobrachem:
Cannot reproduce the issue reported in #185, so this PR simply removes the pin to scipy<=1.12.
Closes #185 

Opened as a draft to see if our CI passes.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -27,7 +27,7 @@ install_requires =
     networkx>=2.6
     numpy>=1.22,!=1.24.0
     pandas>=1.4
-    scipy>=1.8, <=1.12.0
+    scipy>=1.8.0
     seaborn>=0.13
     tensorflow-probability>=0.17
     tqdm>=4.62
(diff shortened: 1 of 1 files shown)


jobrachem:
The issue in the CI does not originate from scipy, but instead from incompatibility of tensorflow probability to numpy 2.0. At first, I could not reproduce the issue locally, because locally, I installed using the `pmc` specs, and pymc currently pins to numpy<2.

The current solution is to pin numpy in liesel to <2.0, too, until the downstream issue is addressed. There's already an issue in the tensorflow probability repo:

- https://github.com/tensorflow/probability/issues/1814

Once this is addressed, we can unpin numpy, since pymc correctly handles its dependency on numpy itself.

PR #202, opened 2024-08-17
Implement easier access to and manipulation of var inputs

jobrachem:
This PR contains only a few lines of changed code, but I think it can drastically improve quality of life while building and manipulating Liesel models.

## Problem statement

I have found the work with variable inputs to be often quite cumbersome. I'll show what I mean in examples.

```python
import liesel.model as lsl
import jax.numpy as jnp
```

### Example 1: Accessing parameters of a distribution

Let's say I want to access the `loc` and `scale` of a variable's distribution, starting from the variable. 

```python
def create_var1():
    loc = lsl.Var(0.0, name="loc")
    scale = lsl.Var(1.0, name="scale")
    y = lsl.obs(jnp.zeros(10), lsl.Dist(tfd.Normal, loc=loc, scale=scale), name="y")
    return y

y1 = create_var1()
```

The `Var.dist_node`'s `kwinputs` attribute does not give me that access. Instead, it returns a dict of `VarValue` nodes. That is node helpful.

```
>>> y1.dist_node.kwinputs
mappingproxy({'loc': VarValue(name="loc_var_value"),
              'scale': VarValue(name="scale_var_value")})
```

To actually get to the `loc` var, I have to call:

```python
>>> y1.dist_node.kwinputs["loc"].var
Var(name="loc")
```

I think this is unnecessarily cumbersome. I have to actually know a lot about Liesel's internals and/or dig in the source code to find what I need. Every time I need to perform this action, I have to look up what to do and extensively test my code in order to be sure that I get it right. It is quite annoying.

### Example 2: Accessing inputs of a calculator

A very similar pattern holds when you use calculators:

```python
def create_var2():
    x = lsl.Var(0.0, name="x")
    y = lsl.Var(lsl.Calc(lambda x: x + 1, x), name="y")
    return y

y2 = create_var2()
```

```python
>>> y2.value_node.inputs
(VarValue(name="x_var_value"),)
```

```python
>>> y2.value_node.inputs[0].var
Var(name="x")
```

## Proposed Solution

I implemented `Node.__getitem__` and `Var.__getitem__` as a remedy. 

The above tasks can now be solved like this:

```python
>>> y1.dist_node["loc"] # access by arg name, here equivalent to y1.dist_node[0]
Var(name="loc")
```

```python
>>> y2[0] # access by index, equivalent to y2.value_node[0]
Var(name="x")
```

### Some details

1. The basic implementation is done in `Node.__getitem__`. 
    a) If it receives an integer, it will essentially look up the searched item in `Node.all_input_nodes()`. This will find all inputs, including positional and keyword inputs.
    b) If it receives a string, it will essentially look up the searched item in `Node.kwinputs`. This will of course find only inputs that are actually keyword inputs.
2. `Var.__getitem__` will defer to its value node. To access inputs to the distribution, users can turn to `Var.dist_node.__getitem__`.

### Setitem

The implementation also provides the possibility to replace inputs via `Node.__setitem__`. Example:

```python
# before
>>> y2[0]
Var(name="x")
>>> y2.value
1.0

# change input
>>> y2[0] = lsl.Var(3.0, name="new_input")

# after
>>> y2[0]
Var(name="new_input")
>>> y2.value
4.0
```

The same works for `Dist`. The implementation is a thin quality-of-life wrapper around the existing `Node.set_inputs()` method.

## Documentation

This is a draft PR for a first discussion. Even if it remains unchanged, documentation has to be added if it ends up being merged.


jobrachem:
Discussion Notes

- There's doubt on whether `lsl.Var.__getitem__` should be able to look up keyword arguments as well as positional arguments when using an integer.
- There's doubt on whether `lsl.Var.__getitem__` should be implemented - a case can be made for just sticking to `lsl.Var.value_node.__getitem__`.

jobrachem:
With the current update I added the following changes:

- Removed the indexing functionality from `Var`
- Restricted integer indexing functionality to access only positional inputs.

This means we now have these possibilities:

## Accessing inputs

### Access named value inputs to a calculator

```python
xvar = Var(2.0, name="xvar")
y = Var(Calc(lambda xarg: xarg + 1.0, xarg=xvar))
y.value_node["xarg"]
```

### Access positional value inputs to a calculator

```python
xvar = Var(2.0, name="xvar")
y = Var(Calc(lambda xarg: xarg + 1.0, xvar))
y.value_node[0]
```

### Access named value inputs to a dist

```python
xvar = Var(2.0, name="xvar")
y = Var(1.0, Dist(tfp.distributions.Normal, loc=xvar, scale=1.0))
y.dist_node["loc"]
```

### Access positional value inputs to a dist

```python
xvar = Var(2.0, name="xvar")
y = Var(1.0, Dist(tfp.distributions.Normal, xvar, scale=1.0))
y.dist_node[0]
```


## Updating inputs

### Updating inputs to a calculator

Note: 

- Works by integer indexing for positional arguments. 
- Can only be used for *updating existing inputs*
- Relies on existing functionality in `Node.set_inputs`, but makes it easier to switch out *one particular* input. `Node.set_inputs` requires users to provide *all* positional and keyword inputs, even if only one input is supposed to be updated.

```python
a = Var(2.0, name="a")
y = Var(Calc(lambda x: x + 1.0, x=a))

b = Var(3.0, name="b")
y.value_node["x"] = b
```

### Updating inputs to a dist

```python
a = Var(2.0, name="a")
y = Var(1.0, Dist(tfp.distributions.Normal, loc=a, scale=1.0))

b = Var(3.0, name="b")
y.dist_node["loc"] = b
```

jobrachem:
Cases to cover in tests:

- What happens when using `del`?
- What happens when trying to assign `None`?
- What happens with variable-args-input calculator functions? e.g. `sum(*args)`
- What happens when we have functions with default arguments and then add or remove an input node that refers to this argument?

jobrachem:
## What happens when using `__del__`?

```python
loc = Var(0.0)
scale = Var(1.0)
x = Calc(lambda loc, scale: loc * scale, loc, scale)
```

```python
del x[0]
```

tries to call

```python
x.__delitem__(0)
```

which results in an `AttributeError`, since we do not implement `Node.__delitem__`. I would say this is expected behavior, because the intention is to provide users with a way to swap out inputs, not to remove inputs without replacement.


## What happens when assigning `None`?

```python
loc = Var(0.0)
scale = Var(1.0)
x = Calc(lambda loc, scale: loc * scale, loc, scale)
```

```python
x[0] = None
```

is equivalent to

```python
x[0] = Value(None)
```

it does not immediately result in an error, since we allow nodes to have a value of `None`. I think this is expected behavior.


## What happens with variable-args-input calculator functions? e.g. `sum(*args)`

The number of inputs gets fixed during node initialization and cannot be changed later through `Node.__setitem__`. I think this is expected behavior. The equivalent is true for variable keyword-inputs functions.

```python
def sum_(*args):
    return sum(args)

x = Calc(sum_, 1.0, 2.0)

assert x.value == pytest.approx(3.0)

x[0] = 2.0
x.update()
assert x.value == pytest.approx(4.0)

with pytest.raises(IndexError):
    x[2] = 3.0
```

## What happens when we have functions with default arguments?

Similar to the above: The inputs get fixed during node initialization and cannot be changed later. I think this is expected behavior.

```python
def sum_(a, b=3.0):
    return a + b

x = Calc(sum_, 1.0)
assert x.value == pytest.approx(4.0)

x[0] = 2.0
x.update()
assert x.value == pytest.approx(5.0)

with pytest.raises(IndexError):
    x[1] = 0.0

with pytest.raises(KeyError):
    x["b"] = 0.0
```

wiep:
@jobrachem, should this already be reviewed or discussed next week?

jobrachem:
@wiep I need to add documentation before this can be reviewed. 

Also, the test actions currently time out, which is not nice for the review process. The timeout seems to be unrelated to our code changes, but the amount of additional time that the tests seem to use is quite drastic.

jobrachem:
See https://github.com/liesel-devs/liesel/issues/213 regarding the tests timing out.

jobrachem:
@wiep @GianmarcoCallegher the merge conflicts are resolved and the tests for circular graph behavior updated :) Ready for review!

jobrachem:
@wiep @GianmarcoCallegher gentle reminder :) Would be meaningful for me to have this soon, since I want to use it in the current semester in teaching.

liesel/model/nodes.py
```diff
@@ -372,6 +372,69 @@ def var(self) -> Var | None:
...
+        input_list: list[Node | Var] = []
+        for input_ in self.inputs:
+            if isinstance(input_, VarValue):
+                if input_.var is None:
+                    raise RuntimeError(f"{input_}.var is None.")
+                input_list.append(input_.var)
+            else:
+                input_list.append(input_)
+
+        return tuple(input_list)
```
GianmarcoCallegher, line 386: This is called by `__get_item__`. We are checking that, if `input.var` is `None`, we raise an error, but even if we try to access to an index whose `Var` is not `None`, thus throwing an error when doing that is allowed only because another input's `Var` is `None`. Does it make sense?

Same with the function `_loc`

GianmarcoCallegher (commented):
Looks good to me. Just a little question

liesel/model/nodes.py
jobrachem: Hi @GianmarcoCallegher, thanks for your comment!
I'm not sure whether I understand the question correctly. Would you go into more detail on what you mean by the sentence „but even if we try to access to an index whose Var is not None, thus throwing an error when doing that is allowed only because another input's Var is None.“? 

Here's what I think:

„try to access to an index whose Var is not None“ In this case, `input_list.append(input_.var)` is called as expected.

GianmarcoCallegher: Maybe a chunk of code will help:

```python
loc = Var(0.0)
scale = Var(1.0)
x = Calc(lambda loc, scale: loc * scale, loc, scale)

x.inputs[0]._var = None

print(x[1])
```

I don't think this should fail since the scale value is not `None`

jobrachem: Ah, thank you for the clarification! I understand now. I would actually argue that the current behavior makes sense, for two reasons:

1. In this scenario, `x.inputs[0]` is a `VarValue` instance that does not have an associated variable. In my opinion, this goes against the purpose of the `VarValue` class (wrapping access to a `Var`'s value), and *should* break things.
2. The scenario is created by accessing a private property. This indicates that users should not handle `VarValue._var`, and cannot expect things to work if they do. Things then don't work, which makes sense.

GianmarcoCallegher: I see your point. In particular the second one. Concerning the first one, I think it makes sense but do you think this is the right place to break things then?

Maybe when building the model would be better in my opinion

jobrachem: Well, at least I don't think it is bad that things break here in this case. I would argue that we don't even need a separate check for this, because of the second point, i.e. the attribute is private and should not even be handled by the user. This is sort of a general thing; I think it is not feasible to explicitly handle the cases in which users could potentially break the software for themselves if they tried to by manipulating private attributes.

GianmarcoCallegher: That's also a valid point. I'd remove it then

jobrachem: When I try to remove it, mypy is unhappy. Would you be ok with leaving the check in there and adding a comment?

GianmarcoCallegher: Sure

jobrachem: Done :)

GianmarcoCallegher:
It's ready to be merged for me

tests/model/test_node.py
```diff
@@ -688,3 +688,274 @@ def test_multivariate_batched_distribution(Dist) -> None:
...
+        assert x[0] is loc
+        assert x[1] is scale
+
+        with pytest.raises(KeyError):
+            x["loc"]
+
+        with pytest.raises(KeyError):
+            x["scale"]
+
+    def test_vars_unnamed(self):
```
wiep, line 728: Do you think more descriptive test names to clearly stating intend could help making sure that we in the future know what should be tested? Could also be a comment.
For example here:  `test_vars_positional_access_to_named_args_raises_error`

liesel/model/nodes.py
```diff
@@ -1050,6 +1115,79 @@ class Var:
     >>> x
     Var(name="x")
 
+    .. rubric:: Accessing inputs
+
+    :class:`.Calc` and :class:`.Dist` objects support access to their inputs via
+    square-bracket syntax. Thus, with a :class:`.Var` object, you can use square bracket
+    indexing on its attributes :attr:`.Var.value_node` and :attr:`.Var.dist_node`.
+    You can access both named and unnamed arguments this way.
```
wiep, line 1123: I think positional arguments is a better/more commonly used term here. Use instead of unnamed.

wiep (approved):
Ready to be merged. Just to comments. One one wording in documentation and the second on more expressive names for tests. Feel free to ignore them if you disargree.

liesel/model/nodes.py
jobrachem: Thanks, I implemented it :)

tests/model/test_node.py
jobrachem: Oh, you are of course absolutely right. The test names were not very descriptive. I improved that now. I also removed some tests in `test_var.py` that have become redundant.

PR #204, opened 2024-08-26
Avoid hardcoded conversion to float32 in GraphBuilder

jobrachem:
There were some places in the GraphBuilder that had a hard-coded conversion to float32. I added an init argument that now makes it possible to avoid this behavior completely. The `lsl.Model` likewise gains the `to_float32` boolean flag as an argument, since if `grow=True`, the model internally uses a graph builder instance for growing.

### DIFF
--- a/liesel/model/model.py
+++ b/liesel/model/model.py
@@ -102,6 +102,12 @@ class GraphBuilder:
     been constructed, some methods of the graph builder are not available anymore,
     because the graph is considered static.
 
+    Parameters
+    ----------
+    to_float32
+        Whether to convert the dtype of the values of the added nodes \
+        from float64 to float32.
+
     See Also
     --------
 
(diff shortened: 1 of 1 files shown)


liesel/model/model.py
```diff
@@ -276,7 +283,7 @@ def _set_missing_names(self) -> GraphBuilder:
         return self
 
     def add(
-        self, *args: Node | Var | GraphBuilder, to_float32: bool = True
+        self, *args: Node | Var | GraphBuilder, to_float32: bool | None = None
```
wiep, line 286: None option is not explained in doc string.

liesel/model/model.py
```diff
@@ -336,7 +346,9 @@ def add(
 
         return self
 
-    def add_groups(self, *groups: Group, to_float32: bool = True) -> GraphBuilder:
+    def add_groups(
+        self, *groups: Group, to_float32: bool | None = None
```
wiep, line 350: None option not mentioned in docstring.

wiep (changes requested):
See the two comments about mentioning the None option in the docstrings.

jobrachem:
@wiep thank you! I added the missing information to the docs.

jobrachem:
See https://github.com/liesel-devs/liesel/issues/213 regarding the tests timing out.

wiep (approved):
perfect; ~~should we wait for #213 before merging?~~

PR #209, opened 2024-08-28
Warn about missing PyGraphviz installation when plotting a lsl.Model

jobrachem:
As the title says. This reduces confusion about why plotted graphs may look chaotic.

### DIFF
--- a/liesel/model/viz.py
+++ b/liesel/model/viz.py
@@ -2,9 +2,13 @@
 Model visualization.
 """
 
+import logging
+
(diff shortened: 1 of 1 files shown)


jobrachem:
@wiep gentle reminder :)

wiep (approved):
PR looks good. One could inform the user about the consequence of not having pygraphbiz. "fallback graph layout" sounds much better than what it actually looks like, i.e., pretty random.

PR #230, opened 2024-12-10
Implement Var.plot_vars and Model.plot_vars

jobrachem:
All of the below applies to `plot_vars` as well as `plot_nodes`.

This PR adds methods to the `lsl.Var` and the `lsl.Model` objects that wrap the referenced plotting functions. The method on `lsl.Model` is inspired by the corresponding `lsl.GraphBuilder.plot_vars` method, which has existed for a long time and I find quite convenient.

The `lsl.Var.plot_vars` brings more novelty, because it is a way to very quickly plot intermediate sub-models. One tricky bit was to deal with the automatic naming: As soon as you initialize a model, Liesel will automatically assign names to unnamed nodes. Because `lsl.plot_vars` works similar to `lsl.GraphBuilder.plot_vars`, this also happens here: A helper model is built, then the plot is created, then the helper model is discarded. The issue is: The previously unnamed nodes keep their new names. I think this is not intended behavior, so the `lsl.Var.plot_vars` method resets the automatically assigned names after plotting. It also emits a logging message that notifies users of this behavior, because I think otherwise it could be confusing.

`lsl.Var.plot_vars` will not work for variables that are already part of a model, because in this case, no intermediate model can be created.

In `lsl.Var.plot_nodes`, the three model nodes (log_prob, log_lik, log_prior) will also show up. I think this is ok. This plot is intended for advanced users anyway, and having the three log prob nodes in there allows users to spot potential issues with the likelihood and prior attribution in their subgraph.

## Example

```python
>>> a = lsl.Var.new_value(1.0)
>>> b = lsl.Var.new_calc(lambda a: a + 1.0, a=a, name="b")
```

```python
>>> a.name
''
```

```python
>>> b.plot_vars()
liesel.model.nodes - INFO - Unnamed variables were temporarily named for plotting. The automatically assigned names are: ['v0']. The names are reset after plotting.
```

image

```python
>>> b.plot_nodes()
liesel.model.nodes - INFO - Unnamed variables were temporarily named for plotting. The automatically assigned names are: ['v0']. The names are reset after plotting.
```

image

Note: In this example, the three model nodes are unconnected, because there is no node with a probability in the minimal example model.

```python
>>> a.name
''
```

### DIFF
--- a/liesel/model/nodes.py
+++ b/liesel/model/nodes.py
@@ -29,6 +29,7 @@ import tensorflow_probability.substrates.numpy.bijectors as nb
 import tensorflow_probability.substrates.numpy.distributions as nd
 
 from ..distributions.nodist import NoDistribution
+from .viz import plot_nodes, plot_vars
 
 if TYPE_CHECKING:
     from .model import Model
@@ -1920,12 +1921,17 @@ class Var:
     def _plot(
         self, which: Literal["vars", "nodes"] = "vars", verbose: bool = True, **kwargs
     ) -> None:
-        if self.model:
+
+        if self.model is not None:
             match which:
                 case "vars":
-                    return self.model.plot_vars()
+                    subgraph = self.model.var_subgraph(self)
+                    return plot_vars(subgraph, **kwargs)
                 case "nodes":
-                    return self.model.plot_nodes()
+                    self_nodes = [self.value_node, self.dist_node, self.var_value_node]
+                    filtered_nodes = [nd for nd in self_nodes if nd is not None]
+                    subgraph = self.model.node_subgraph(*filtered_nodes)
+                    return plot_nodes(subgraph, **kwargs)
 
         from liesel.model import GraphBuilder
 
@@ -1952,9 +1958,14 @@ class Var:
 
         match which:
             case "vars":
-                model.plot_vars(**kwargs)
+                subgraph = model.var_subgraph(self)
+                plot_vars(subgraph, **kwargs)
             case "nodes":
-                model.plot_nodes(**kwargs)
+                self_nodes = [self.value_node, self.dist_node, self.var_value_node]
+                self_nodes = [nd for nd in self_nodes if nd is not None]
+                filtered_nodes = [nd for nd in self_nodes if nd is not None]
+                subgraph = model.node_subgraph(*filtered_nodes)
+                plot_nodes(subgraph, **kwargs)
 
         model.pop_nodes_and_vars()
 
--- a/liesel/model/nodes.py
+++ b/liesel/model/nodes.py
@@ -1925,12 +1925,12 @@ class Var:
         if self.model is not None:
             match which:
                 case "vars":
-                    subgraph = self.model.var_subgraph(self)
+                    subgraph = self.model.var_parental_subgraph(self)
                     return plot_vars(subgraph, **kwargs)
                 case "nodes":
                     self_nodes = [self.value_node, self.dist_node, self.var_value_node]
                     filtered_nodes = [nd for nd in self_nodes if nd is not None]
-                    subgraph = self.model.node_subgraph(*filtered_nodes)
+                    subgraph = self.model.node_parental_subgraph(*filtered_nodes)
                     return plot_nodes(subgraph, **kwargs)
 
         from liesel.model import GraphBuilder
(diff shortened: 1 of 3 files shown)


wiep:
> In lsl.Var.plot_nodes, the three model nodes (log_prob, log_lik, log_prior) will also show up. I think this is ok. This plot is intended for advanced users anyway, and having the three log prob nodes in there allows users to spot potential issues with the likelihood and prior attribution in their subgraph.

I think this is rather a feature, because when your var has an associated distribution the user can varify that it contributes to the model_log_prob / likelihood / prior node.

jobrachem:
Maybe the necessity to build an intermediate model can be removed by using `lsl.Model._build_graph` directly. This would be very nice, because it would mean that you can also simply plot the subgraph of a variable that *is* part of a model.

jobrachem:
> > In lsl.Var.plot_nodes, the three model nodes (log_prob, log_lik, log_prior) will also show up. I think this is ok. This plot is intended for advanced users anyway, and having the three log prob nodes in there allows users to spot potential issues with the likelihood and prior attribution in their subgraph.
> 
> I think this is rather a feature, because when your var has an associated distribution the user can varify that it contributes to the model_log_prob / likelihood / prior node.

I agree 😊

jobrachem:
> > > In lsl.Var.plot_nodes, the three model nodes (log_prob, log_lik, log_prior) will also show up. I think this is ok. This plot is intended for advanced users anyway, and having the three log prob nodes in there allows users to spot potential issues with the likelihood and prior attribution in their subgraph.
> > 
> > 
> > I think this is rather a feature, because when your var has an associated distribution the user can varify that it contributes to the model_log_prob / likelihood / prior node.
> 
> I agree 😊

I thought a bit more on it, and I now think the behavior is not all that useful. The model nodes we see on the graph above are not model nodes from the "true" model, but from an auxiliary model that is created just to construct the graph for plotting. So connections to these nodes are not diagnostic of the presence or absence of problems in the "true" model that includes child nodes.

I added some commits with the following updates:

1. Calling Var.plot_vars() and Var.plot_nodes() will now also work as expected when the Var is already part of a model. In the process, I added the methods `Model.var_subgraph()` and `Model.node_subgraph`.
2. The model nodes are not part of the plot produced by Var.plot_nodes anymore.

jobrachem:
A more involved example:

```python
import liesel.model as lsl

mu = lsl.Var.new_param(1.0, lsl.Dist(tfd.Normal, loc=0.0, scale=1.0), name="mu")
sigma = lsl.Var.new_param(
    1.0,
    lsl.Dist(
        tfd.InverseGamma,
        concentration=lsl.Var.new_value(2.0),
        scale=lsl.Var.new_value(0.5),
    ),
    name="sigma",
)
y = lsl.Var(1.0, lsl.Dist(tfd.Normal, loc=mu, scale=sigma), name="y")

y.plot_vars()
```

grafik


```python
model = lsl.Model([y])
y.plot_vars()
```

grafik

```python
y.plot_nodes()
```

grafik

wiep:
Do the new function *_subgraph show the subgraph of the  node and its parents? If so, maybe we should not use the very general term subgraph here but rather parental_subgraph or something similar that is more descriptive. What do you think.


PS: I agree with what you wrote above about the meaningless temporary model.

jobrachem:
Hey @wiep, fair point. I changed the names and added a line of docstring to each of the two methods.

jobrachem:
@wiep can we merge this?

wiep:
I'd say yes but I did not review the code very carefully.

jobrachem:
Okay, I suppose it can still wait until the next weekly, then we can maybe have another chat about it.

liesel/model/nodes.py
```diff
@@ -1949,6 +1959,177 @@ def weak(self) -> bool:
...
+    ) -> None:
+
+        if self.model is not None:
+            match which:
+                case "vars":
+                    subgraph = self.model.var_parental_subgraph(self)
+                    return plot_vars(subgraph, **kwargs)
+                case "nodes":
+                    self_nodes = [self.value_node, self.dist_node, self.var_value_node]
+                    filtered_nodes = [nd for nd in self_nodes if nd is not None]
```
GianmarcoCallegher, line 1973: Aren't these two lines doing the same?

liesel/model/nodes.py
```diff
@@ -1949,6 +1959,177 @@ def weak(self) -> bool:
...
+                "Unnamed variables were temporarily named for plotting. The"
+                f" automatically assigned names are: {var_names}. The names are reset"
+                " after plotting."
+            )
+        if node_names and verbose:
+            logger.info(
+                "Unnamed nodes were temporarily named for plottingThe automatically"
+                f" assigned names are: {node_names}. The names are reset after"
+                " plotting."
+            )
```
GianmarcoCallegher, line 1996: I think this is extremely useful, but if there are too many variables without a name, the list `var_names' might be a bit too long. Maybe we could add a `verbose` parameter and print them only if it is `True`

jobrachem: Sounds like  good idea to me! :)

jobrachem: I made three slight changes: 

1. If `verbose=False`, the list of names is omitted, but the general message is still logged. I think this might be important to avoid confusion.
2. I switched the default from `verbose=True` to `verbose=False`.
3. I added a docstring entry for the `verbose` argument.

jobrachem:
Hey @GianmarcoCallegher I removed the duplicate line and made a small update to the `verbose` functionality. If you are ok with that, I think we can merge :)

PR #232, opened 2024-12-12
Weaken jax version pin to <0.5.0, closes #214

### LINKED ISSUE #214: Remove upper pin for jax
Jax was pinned to solve this: #213

jobrachem:
The issue in our code originated in the `liesel.goose.optim` module, specifically in the callback that updates the progress bar.

In the callback, the current loss values were accessed like this:

```python
i = val["while_i"]
loss_train = val["history"]["loss_train"][i]
loss_validation = val["history"]["loss_validation"][i]
```

Some experimentation revealed that this pattern was the problem. Introducing dedicated `"current_loss_train"` and `"current_loss_validation"` carry values solved the issue:

```python
loss_train = val["current_loss_train"]
loss_validation = val["current_loss_validation"]
```

While I was at it, I fixed two other issues with the progress bar:

1. There was a bug that prevented the argument `progress_bar=False` from taking effect, i.e. the progress bar couldn't be turned off. This is fixed now.
2. The progress bar displayed `max_iter-1` as the total number of iterations. This is fixed now.
3. When a user defines a gs.Stopper(patience=10) and does not pass a model_validation, they will have reason to expect early stopping to be performed based on the training model. Previously, this belief was false. Now this belief is true.

### DIFF
--- a/liesel/goose/optim.py
+++ b/liesel/goose/optim.py
@@ -333,7 +333,7 @@ def optim_flat(
     >>> stopper = gs.Stopper(max_iter=1000, patience=10, atol=0.01)
     >>> result = gs.optim_flat(model, params=["coef"], stopper=stopper)
     >>> {name: jnp.round(value, 2) for name, value in result.position.items()}
-    {'coef': Array([0.52, 1.29], dtype=float32)}
+    {'coef': Array([0.52, 1.27], dtype=float32)}
 
     We can now, for example, use ``result.model_state`` in
     :meth:`.EngineBuilder.set_initial_values` to implement a "warm start" of MCMC
@@ -455,6 +455,8 @@ def optim_flat(
     init_val["history"] = history
     init_val["position"] = position
     init_val["opt_state"] = optimizer.init(position)
+    init_val["current_loss_train"] = history["loss_train"][0]
+    init_val["current_loss_validation"] = history["loss_validation"][0]
     init_val["key"] = jax.random.PRNGKey(batch_seed)
     init_val["model_state_train"] = model_train.state
     init_val["model_state_validation"] = model_validation.state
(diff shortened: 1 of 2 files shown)


liesel/goose/optim.py
```diff
@@ -360,7 +360,6 @@ def optim_flat(
     user_patience = stopper.patience
     if model_validation is None:
         model_validation = model_train
-        stopper.patience = stopper.max_iter
```
wiep, line 363: This change seems to be unrelated to the issue. Should we keep it and mention it in the PR?

wiep (approved):
The changes look sensible to me, although I do not understand why the old method did cause issues.

liesel/goose/optim.py
jobrachem: Thanks for catching that @wiep! What happens here is that, if no `model_validation` was set explicitly, early stopping is turned off.

When a user defines a `gs.Stopper(patience=10)` and does not pass a `model_validation`, they will have reason to expect early stopping to be performed based on the training model. Previously, this belief was false. Removing this line makes this belief true.

I'm in favor of keeping this change and mentioning it.

jobrachem: I added a mention of it to the PR description.

wiep: Wouldn't the expectation be that if there's no validation model supplied no validation and no early stopping is performed? If I want to have the training model to be used as a validation model I would explicitly supply it twice to the algorithm: once as training model and once as validation model.

jobrachem: That was indeed the reasoning for the initial syntax. But the point about the stopper still stands – and I don't think it is necessarily true that without a validation model, you always want to run the maximum number of iterations. Even without a validation model, I would want to stop if there is no movement in the loss anymore.

wiep: I guess I want the user to be explicit that they want to use the same data for training and validation. I am not sure if there is an expectation that the data is automatically split. Anyway, I guess that is a different issue.

wiep:
JAX released 0.5.0 which has some potential breaking changes in PRNG. I think for us might be of interest that the 'high-variance' test procedures might fail without updated values. https://github.com/jax-ml/jax/discussions/18480.

wiep:
On my computer, the new Jax version breaks multiple tests. I pinned jax in this PR to <0.5.0 and will create a new issue for the PRNG problem. Thus, we can merge this and focus on fixing the tests in a new PR.

PR #231, opened 2024-12-10
Adds edge to plot_vars if input is used in value while var has distribution

wiep:
Fixes #217.

Adds an edge if input is used in value of a node that has also a distribution. Introduces a red edge if input is used in value and dist node.

Changes the created by the code
```py
import liesel.model as lsl
import jax.numpy as jnp
import tensorflow_probability.substrates.jax.distributions as tfd

# create a model
x = lsl.Var(jnp.arange(4).astype(float), name="x")
y = lsl.Var(jnp.arange(4).astype(float), name="y")
batch_index = lsl.Var(
    jnp.array([0, 1]),
    name="batch_index",
)
y_sel_dist = lsl.Dist(lambda x, idx: tfd.Normal(x[idx], 1.0), x, batch_index)
y_sel_val = lsl.Calc(lambda y, idx: y[idx], y, batch_index)
y_sel_val.value
y_sel = lsl.Var(
    y_sel_val,
    y_sel_dist,
    "y_sel",
)

model = lsl.GraphBuilder().add(y_sel_dist).build_model()
lsl.plot_vars(model)
```
from
old
to
new

PR #235, opened 2024-12-15
Allow transformation of weak vars

jobrachem:
So far, we did not allow transformation of weak vars in `Var.transform`. I have a usecase for that right now, and I don't think anything really speaks against allowing it. This PR makes two small changes that enable the functionality.

If you transform a weak var, it gets inserted in between the old value node and the transformed variable:

```python
import liesel.model as lsl
import jax.numpy as jnp

tau = lsl.Var.new_param(10.0, name="tau")
tau_sqrt = lsl.Var.new_calc(jnp.sqrt, tau, name="tau_sqrt")
log_tau_sqrt = tau_sqrt.transform(tfb.Exp())
model = lsl.Model([tau_sqrt])
lsl.plot_vars(model)
```

image

### DIFF
--- a/liesel/model/nodes.py
+++ b/liesel/model/nodes.py
@@ -756,8 +756,9 @@ class Calc(Node):
     ):
         super().__init__(*inputs, **kwinputs, _name=_name, _needs_seed=_needs_seed)
         self._function = function
+        self.update_on_init = update_on_init
 
-        if update_on_init:
+        if self.update_on_init:
             try:
                 self.update()
             except Exception as e:
@@ -1971,11 +1972,43 @@ def _transform_var_with_bijector_instance(var: Var, bijector_inst: jb.Bijector) 
 
     transformed_dist.per_obs = var.dist_node.per_obs
 
-    transformed_var = Var(
-        bijector_inv.forward(var.value),
-        transformed_dist,
-        name=f"{var.name}_transformed",
-    )
+    if var.weak:
+        try:
+            value_function = var.value_node.function  # type: ignore
+        except AttributeError as e:
+            raise AttributeError(
+                "Trying to transform a weak variable without calculator node."
+            ) from e
+
+        def forward(*args, **kwargs):
+            return bijector_inv.forward(value_function(*args, **kwargs))
+
+        value_inputs = var.value_node.inputs
+        value_kwinputs = var.value_node.kwinputs
+        value_node_needs_seed = var.value_node.needs_seed
+        try:
+            value_node_upadte_on_init = var.value_node.update_on_init  # type: ignore
+        except AttributeError as e:
+            raise e
+
+        transformed_var = Var(
+            Calc(
+                forward,
+                *value_inputs,
+                _name="",
+                _needs_seed=value_node_needs_seed,
+                update_on_init=value_node_upadte_on_init,
+                **value_kwinputs,
+            ),
+            transformed_dist,
+            name=f"{var.name}_transformed",
+        )
+    else:
+        transformed_var = Var(
+            bijector_inv.forward(var.value),
+            transformed_dist,
+            name=f"{var.name}_transformed",
+        )
 
     var.value_node = Calc(bijector_inst.forward, transformed_var)
     return transformed_var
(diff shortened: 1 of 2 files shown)


wiep:
What happens if the weak variable has a distribution associated with it?

jobrachem:
Great point @wiep, thanks! The case was not covered. On application, the current value of the variable that is being transformed would have been fixed.

I added a case and a test now to cover this.

Currently, the model graph looks a bit weird as a result:

```python
import liesel.model as lsl
import jax.numpy as jnp

x = lsl.Var.new_value(jnp.linspace(0.1, 2, 5), name="all_x")
batch_index = lsl.Var.new_value(1, name="index")

x_batched = lsl.Var(
    value=lsl.Calc(lambda i, x: x[i], batch_index, x),
    distribution=lsl.Dist(tfd.Normal, loc=0.0, scale=1.0),
    name="x_batched",
)

x_batched_transformed = x_batched.transform(tfb.Exp())
model = lsl.Model([x_batched])
```

```python
lsl.plot_vars(model)
```

image


The node graph seems to indicate that things are connected like they should be:

```python
lsl.plot_nodes(model)
```

image


But https://github.com/liesel-devs/liesel/pull/231 will fix this:

image


Also for the case that the index is used in both the value and the distribution:

```python
x = lsl.Var.new_value(jnp.linspace(0.1, 2, 5), name="all_x")
m = lsl.Var.new_param(jnp.arange(5).astype(float), name="m")
batch_index = lsl.Var.new_value(1, name="index")

x_batched = lsl.Var(
    value=lsl.Calc(lambda i, x: x[i], batch_index, x),
    distribution=lsl.Dist(lambda i, m: tfd.Normal(m[i], 1.0), batch_index, m),
    name="x_batched",
)

x_batched_transformed = x_batched.transform(tfb.Exp())
model = lsl.Model([x_batched])
lsl.plot_vars(model)
```

image

wiep:
nice :-) 

Looks like you're running with the same use case into the same issue that I had with plotting.

jobrachem:
@wiep this PR is rebased, and the plotting issue is gone :)

wiep (approved):
thanks. lgtm.

jobrachem:
@wiep @GianmarcoCallegher as promised, some more details on my usecase. Imagine you have a variance parameter, and want to define a regression model for this parameter. For the time being, you are happy with defining the model and having the parameter available for further use.

Then, at a later point, you want to transform the model to actually be defined on the log scale. Maybe you also want to easily compare log and softplus transformation, etc. If you can simply transform the existing variance parameter, this is very easy.

```python
import liesel.model as lsl
import jax.numpy as jnp

# initial model definition
x = lsl.Var.new_obs(jnp.array([[1.0]]), name="x")
b = lsl.Var.new_param(jnp.ones(1), name="b")
s = lsl.Var.new_calc(jnp.dot, x, b, name="s") # actually a preliminary model
y = lsl.Var.new_obs(jnp.zeros(1), lsl.Dist(tfd.Normal, loc=0.0, scale=s), name="y")

# Apply bijector later
log_s = s.transform(tfb.Exp())

lsl.plot_vars(lsl.Model([s]))
```

image


The alternative would be to use the new square bracket access to swap out the dist input with a newly defined calculator variable. This is not too bad and requires only one extra line of code – but it also requirs knowledge about how exactly the parameter `s` is being used downstream. Simply calling transform eliminates this need for knowledge.

```python
# initial model definition
x = lsl.Var.new_obs(jnp.array([[1.0]]), name="x")
b = lsl.Var.new_param(jnp.ones(1), name="b")
s = lsl.Var.new_calc(jnp.dot, x, b, name="s") # actually a preliminary model
y = lsl.Var.new_obs(jnp.zeros(1), lsl.Dist(tfd.Normal, loc=0.0, scale=s), name="y")

# Apply bijector later
log_s = lsl.Var.new_calc(jnp.exp, s, name="log_s")
y.dist_node["scale"] = log_s

lsl.plot_vars(lsl.Model([y]))
```

jobrachem:
@wiep turns out, I commited a thinko with my usecase. I would want the two following ways of doing things to be equivalent:

### Target

```python
import liesel.model as lsl
import jax.numpy as jnp

x = lsl.Var.new_obs(jnp.array([[1.0]]), name="x")
b = lsl.Var.new_param(jnp.ones(1), name="b")
s_transformed = lsl.Var.new_calc(jnp.dot, x, b, name="s_transformed") # actually a preliminary model
y = lsl.Var.new_obs(jnp.zeros(1), lsl.Dist(tfd.Normal, loc=0.0, scale=s_transformed), name="y")

# Apply bijector later
s = lsl.Var.new_calc(jnp.exp, s_transformed, name="s")

print(s_transformed.value)
print(s.value)
```

```
[1.]
[2.7182817]
```


image


### Implemented

```python
import liesel.model as lsl
import jax.numpy as jnp

# initial model definition
x = lsl.Var.new_obs(jnp.array([[1.0]]), name="x")
b = lsl.Var.new_param(jnp.ones(1), name="b")
s = lsl.Var.new_calc(jnp.dot, x, b, name="s") # actually a preliminary model
y = lsl.Var.new_obs(jnp.zeros(1), lsl.Dist(tfd.Normal, loc=0.0, scale=s), name="y")

# Apply bijector later
log_s = s.transform(tfb.Exp())
print(log_s.value)
print(s.value)
```

```
[0.]
[1.]
```

image


### The important difference

In the first version, the transformed variable and its output are

```
s_transformed = x @ b
s = exp(s_transformed)
```


In the second version, we have

```
s_transformed = ln(x @ b)
s = exp(s_transformed)
```

This second version will often fail. In this case, for example, as soon as `x @ b` becomes negative.

The first version would be nicer, but not consistent with the current API of the `Var.transform` method.

PR #241, opened 2025-02-14
Liesel version in result object

GianmarcoCallegher:


GianmarcoCallegher:
Would you store the output `pip list` simply as a string?

jobrachem:
Hey @GianmarcoCallegher, I would be hesitant to add a dependency for this. Does it even make sense to store the git commit if liesel is installed via pip? I would expect an installation via pip for most users. I think to keep things simple, it would be good to limit this to storing the liesel version, like in the current commit. 

Storing the output of pip list may be nice, if it works easily across different setups and runtime environments. In that case, I think storing it as a string would be sufficient. If it causes implementation complications, I would not do it. In any case, I wonder if storing pip list is not a little "overreach" on the liesel side, i.e. doing more than necessary and expected.

GianmarcoCallegher:
> Hey @GianmarcoCallegher, I would be hesitant to add a dependency for this. Does it even make sense to store the git commit if liesel is installed via pip? I would expect an installation via pip for most users. I think to keep things simple, it would be good to limit this to storing the liesel version, like in the current commit.
> 
> Storing the output of pip list may be nice, if it works easily across different setups and runtime environments. In that case, I think storing it as a string would be sufficient. If it causes implementation complications, I would not do it. In any case, I wonder if storing pip list is not a little "overreach" on the liesel side, i.e. doing more than necessary and expected.

I agree regarding the first point. I showed how this could be achieved just because it was part of the description of the original issue.

The second point could be achieved in this way:

```python
import sys
import subprocess

result = subprocess.run([sys.executable, "-m", "pip", "list"], capture_output=True, text=True)
pip_list_output = result.stdout

print(pip_list_output)  # Print to verify
```

This should work in most environments like Windows, macOS, and Linux, but I am not sure it will work everywhere. Considering that this is not such a fundamental feature, I would not introduce it knowing it might cause potential issues

jobrachem:
Sounds reasonable to me.

GianmarcoCallegher:
Then, it can reviewed

jobrachem (approved):

PR #243, opened 2025-03-29
Allow distributions in Var.new_calc

jobrachem:
Small PR that allows users to specify a distribution in `Var.new_calc`. I originally thought this was unnecessary, but there are some edge cases when it might be useful. 

For example, in a copula model, you might have two strong variables with distributions representing the margins, and a weak calculator variable on the u-level after probability transformation. This weak caculator may receive the copula distribution to complete the setup.

### DIFF
--- a/liesel/model/nodes.py
+++ b/liesel/model/nodes.py
@@ -1333,6 +1333,7 @@ class Var:
         cls,
         function: Callable[..., Any],
         *inputs: Any,
+        distribution: Dist | None = None,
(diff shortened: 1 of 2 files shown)


GianmarcoCallegher (commented):
It looks good, but perhaps we should add a tiny test?

jobrachem:
Sure, I added a test :)

GianmarcoCallegher (approved):
Awesome

Issue #42, opened 2023-03-03
Let Groups have an an opinion about how their member Params can be sampled.


jobrachem:
DRAFT ... to be elaborated.

```python
class Group:
    ...
    def kernels(self) -> list[Kernel]:
        ...
```

wiep:
Would it make sense to include like the inference interface we have for vars also for groups?

PR #244, opened 2025-03-29
Introduce argument 'mcmc_kernel' to lsl.Var

jobrachem:
## Motivation

Often, when I build a liesel model and a corresponding MCMC algorithm, I find that, when it is time to define my MCMC kernels, I cannot remember all of the names of the variables I want to sample. As a result, I jump back and forth between the `gs.EngineBuilder` code, and the model definition code, looking up and copy-pasting variable names for kernel initialization. 

Also, when I add or remove a variable, I always have to update two places in the code: Model definition, and MCMC algorithm.

This feels like a lot of empty busy-work. I would like more convenience.

## This PR

This is a small PR that adds an init argument and attribute `mcmc_kernel` to lsl.Var and a method `mcmc_kernels` to lsl.Model. 

- The argument allows you to pass a `gs.Kernel` class or instance, which will afterwards be available in the `Var.mcmc_kernel` property. 
    - The property always returns a kernel *instance*. If a class object was passed as the argument, the class will be instantiated as `Kernel([self.name])`, i.e. with the variable name as the sole position key and default arguments otherwise.
- The `Model.mcmc_kernels` method returns a dictionary of all kernels found at variables in the model.

The purpose is convenience: When defining a Liesel model, I can specify the desired MCMC kernel right where I define the variable. This makes it easy to in- or exclude individual variables from the model. 

## Pattern example

```python
mu = Var(
    value=0.0,
    distribution=Dist(tfd.Normal, loc=0.0, scale=1.0),
    name="mu",
    mcmc_kernel=gs.NUTSKernel,
)

sigma = Var(
    value=1.0,
    distribution=Dist(tfd.InverseGamma, concentration=0.01, scale=0.01),
    name="sigma",
    mcmc_kernel=gs.HMCKernel(["sigma"], da_target_accept=0.6)
)

model = Model([mu])

eb = gs.EngineBuilder()

for kernel in model.mcmc_kernels().values():
    eb.add_kernel()
```

## Connection between model and estimation

This PR introduces a crack into the strict separation of model and estimation that we otherwise have in the Liesel library. However, I think that the gain in user-friendlyness is worth it.

Something similar was long ago hinted at here: 

- #42

I think this PR may be considered as closing that issue.

## Transformed variables

If a variable with an mcmc kernel is transformed, the kernel is removed without replacement. Users are warned. The intended pattern for adding mcmc kernels to transformed variables is:

```python
a = lsl.Var(0.0, name="a")
b = a.transform(tfb.Exp()) # If any kernel was defined, it will be removed here
b.mcmc_kernel = gs.NUTSKernel
```

### DIFF
--- a/liesel/model/model.py
+++ b/liesel/model/model.py
@@ -11,7 +11,7 @@ from collections import Counter
 from collections.abc import Iterable
 from copy import deepcopy
 from types import MappingProxyType
-from typing import IO, Any, Literal, TypeVar
+from typing import IO, TYPE_CHECKING, Any, Literal, TypeVar
 
 import dill
 import jax
@@ -39,6 +39,9 @@ from .nodes import (
     is_bijector_class,
 )
 from .viz import plot_nodes, plot_vars
+
+if TYPE_CHECKING:
+    from ..goose.kernel import Kernel
 
 __all__ = ["GraphBuilder", "Model", "load_model", "save_model"]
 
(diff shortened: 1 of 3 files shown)


jobrachem:
Closing this PR in favor of #253

PR #246, opened 2025-03-29
Add Model.predict and Var.predict

### LINKED ISSUE #54: Predictions in Liesel
This issue serves as a reminder and a tracker for the discussion about convenience functionality for predictions in Liesel. Currently, we do not really offer much in that regard.

I propose that we could let certain `lsl.Group` child-classes know about how they want to be used for prediction, for example:

```python
class PSpline(lsl.Group):

    def predict(self, samples, new_data) -> Array:
    
        coef_samples = samples[self.coef.name]
        basis_matrix = self.basis_matrix.eval_basis_funs(new_data)
        smooth = np.tensordot(
            basis_matrix, coef_samples, axes=([1], [-1])
        )

        return np.moveaxis(smooth, 0, -1) # makes axes fit to samples
```

What's more, @wiep mentioned that it may be nice to utilize the model graph for predictions.

jobrachem:
This PR adds predictions to Liesel. The predictions are based on updating the model graph with new values repeatedly based on a dictionary of samples. For this, I use `jax.vmap`.

In the process, I added the methods `Model.extract_position()` and `Model.update_state()`, which basically mirror the corresponding methods in the `gs.LieselInterface`.

### Some features:

- Users can supply a dictionary of `newdata` at which to evaluate predictions. If they don't, `predict` uses the current values.
- Users can choose to evaluate predictions only for a subset of variables, indicated by a list of variable names. In this case, a `Model.parental_submodel()` for the selected variables is created. This avoids unnecessary computations and makes it possible to, for example in a regression setting, evaluate predictions on grids for covariates without having to supply response values of fitting shape.
- `Model.predict()` returns a dictionary of predictions for a) all variables in the model (if none are selected specifically), or b) the selected variables and/or nodes.
- `Var.predict()` returns a single array of predictions for this specific variable.

### Demo Notebook:

[test_predictions.ipynb.zip](https://github.com/user-attachments/files/19522955/test_predictions.ipynb.zip)

- Closes #54

### DIFF
--- a/liesel/goose/interface.py
+++ b/liesel/goose/interface.py
@@ -280,16 +280,7 @@ class LieselInterface:
         model_state
             A dictionary of node names and their corresponding :class:`.NodeState`.
         """
-        position = {}
-
-        for key in position_keys:
-            try:
-                position[key] = model_state[key].value
-            except KeyError:
-                node_key = self._model.vars[key].value_node.name
-                position[key] = model_state[node_key].value
-
-        return Position(position)
+        return self._model.extract_position(position_keys, model_state)  # type: ignore
 
     def update_state(self, position: Position, model_state: ModelState) -> ModelState:
         """
@@ -309,23 +300,7 @@ class LieselInterface:
         ``position``. If you supply a ``model_state`` with outdated nodes, these nodes
         and their outputs will not be updated.
         """
-
-        # sets all outdated flags in the model state to false
-        # this is required to make the function jittable
-
-        self._model.state = model_state
-
-        for node in self._model.nodes.values():
-            node._outdated = False
-
-        for key, value in position.items():
-            try:
-                self._model.nodes[key].value = value  # type: ignore  # data node
-            except KeyError:
-                self._model.vars[key].value = value
-
-        self._model.update()
-        return self._model.state
+        return self._model.update_state(position, model_state)
 
     def log_prob(self, model_state: ModelState) -> float:
         """
--- a/liesel/model/model.py
+++ b/liesel/model/model.py
@@ -1234,6 +1234,29 @@ class Model:
         subgraph = self.var_graph.subgraph(nodes_to_include)
         return subgraph
 
+    def parental_submodel(self, *of: Var | Node) -> Model:
+        """
+        Returns a new model that consists only of the given variables and nodes and \
+        their parent variables and nodes. The new model contains copies of these \
+        variables and nodes.
+        """
+        nodes_to_include = set()
+
+        for node in of:
+            if isinstance(node, Var):
+                nodes_to_include.update(nx.ancestors(self.var_graph, node))
+            else:
+                nodes_to_include.update(nx.ancestors(self.node_graph, node))
+            nodes_to_include.add(node)
+
+        copy_of_nodes_to_include = deepcopy(nodes_to_include)
+
+        for node in copy_of_nodes_to_include:
+            if hasattr(node, "_unset_model"):
+                node._unset_model()
+
+        return Model(list(copy_of_nodes_to_include))
+
     @property
     def log_lik(self) -> Array:
         """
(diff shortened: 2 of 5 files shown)


jobrachem:
Response to copilot's comments:

- Copilot correctly points out that a specific shape is expected for the `samples` argument, i.e. `(chains, draws, ...)`. This is relevant for comment `tests/model/test_model.py:461` and `liesel/model/model.py:1701`. The requirement is stated in the documentation. Adding an explicit shape check seems a little difficult to be, because the only thing one may reasonably check is whether there are >=2 dimensions for each array in `samples`. Do you think that would provide benefit over the current behavior?
- Copilot correctly points out that using `deepcopy` may impact performance in large models. Deepcopy is called when a submodel is created, but not if `predict=None` (the default). I am open to suggestions for how to avoid deepcopy, but I am not shure whether it can be avoided if we want to set up a submodel without destroying the current one. (`liesel/model/model.py:1252`)
- `liesel/model/nodes.py:2110`: Fair point. Actually, there's an even better way: Using the `@in_model_method` decorator. This is done now.

jobrachem:
Regarding the shapes of posterior samples: I could imagine adding something like an `n_batching_dims: int = 2` argument. Then the function could do some internal reshaping (combining the batching dimensions), use a single call to `jax.vmap`, and then reshape back to the original batching dimensions.

liesel/model/model.py
```diff
@@ -1533,6 +1556,150 @@ def plot_nodes(
             prog=prog,
         )
 
+    def extract_position(
```
wiep, line 1559: These two methods seem to be almost duplicating the functionality in `GooseModel.extract_position` and `GooseModel.update_state`. Does it make sense to remove this duplication and call these the new functions from `GooseModel`?

liesel/model/nodes.py
```diff
@@ -2087,6 +2087,31 @@ def plot_vars(
             prog=prog,
         )
 
+    @in_model_method
+    def predict(
```
wiep, line 2091: I'm wondering about the intended use of this function. Aside from the tests, I don't see any other places where it's being called. Is there a scenario where this functionality is needed independently of making a model prediction?

From my perspective, prediction generally only makes sense within the context of a model. This function feels more like a convenience wrapper. But is it something we expect to be a common use case? trying to understand whether it justifies being a standalone utility.

wiep (commented):
PR looks clean! Just one comment to avoid code duplication and one with need for clarification.

liesel/model/nodes.py
jobrachem: This is indeed a convenience wrapper. It enables programmatic access to prediction for individual variables, such that you don't have to type out their name. I find it especially useful, because in some contexts, there is no autocompletion for dictionary keys.

Here's an example (granted, it is a little exagerrated):

```python
predictions = model.predict(
    samples,
    predict=["long_complicated_name_that_i_dont_remember"]
) # model method, returns a dict

predicted_values = predictions["long_complicated_name_that_i_dont_remember"]
```


vs.

```python
predicted_values = var.predict(samples) # variable method, returns the array directly
```

liesel/model/model.py
jobrachem: @wiep It is implemented :)

liesel/goose/interface.py
```diff
@@ -280,16 +280,7 @@ def extract_position(
...
-
-        for key in position_keys:
-            try:
-                position[key] = model_state[key].value
-            except KeyError:
-                node_key = self._model.vars[key].value_node.name
-                position[key] = model_state[node_key].value
-
-        return Position(position)
+        return self._model.extract_position(position_keys, model_state)  # type: ignore
```
wiep, line 283: The changes in this file are covered by existing tests in https://github.com/liesel-devs/liesel/blob/c9b3822a40bd3f74cc25361ac013653f36c0614a/tests/goose/test_interface_liesel.py#L23.

wiep (approved):
May be merged. thanks for the work.

PR #248, opened 2025-03-29
Add Model.sample and Var.sample

jobrachem:
This PR uses `Model.extract_position()` and `Model.update_state()` as introduced in #246 extensively. It should not be reviewed before #246 is merged.

This PR adds `Model.sample` and `Var.sample`. It builds upon `Model.simulate`, but expands the functionality and offers a different API. Core summary:

- While `.simulate` updated the model state in place with simulated values, `.sample` leaves the current model unchanged and instead returns a dictionary of sampled values.
- While `.simulate` allowed only for drawing one sample (i.e. updating the model state once with random draws), `.sample` allows you to specify a desired batch shape of samples. That makes it very convenient for prior predictive simulations. Internally, it uses `jax.vmap`.
- `.sample` accepts an argument `posterior_samples`. If you supply a dictionary of `posterior_samples`, the method will conduct posterior predictive samples.
- `Model.sample` will sample from all variables with distributions. 
- `Var.sample` will sample from all variables with distributions in the parental submodel defined by `Var`.
- `.sample` plays nicely with predictions as proposed in #246. I.e., you can pass the output of `.sample` directly to `Model.predict()` and `Var.predict()`.
- `.simulate`'s `skip` argument is renamed to `fixed` in `.sample`

- [ ] I think `.sample` could replace `simulate` -- also in the spirit of "there should be one - and preferably only one - obvious way to do it". Currently however, the PR does not include deprecation for `.simulate`.

## Minimal illustration

```python

import liesel.model as lsl
import jax.numpy as jnp
import jax

n = 100

X = jax.random.uniform(jax.random.PRNGKey(1), (n,2))
b = jnp.array([1.0, 2.0])
eps = jax.random.normal(jax.random.PRNGKey(2), (n,))

y = X @ b + eps

Xvar = lsl.Var.new_obs(X, name="X")
bvar = lsl.Var.new_param(b, lsl.Dist(tfd.Normal, loc=0.0, scale=1.0), name="b")
loc = lsl.Var.new_calc(jnp.dot, Xvar, bvar, name="loc")

scale = lsl.Var.new_param(
    jnp.array([1.0]), 
    distribution=lsl.Dist(tfd.Exponential, rate=1.0),
    name="scale"
)
scale.transform(tfb.Exp())

yvar = lsl.Var.new_obs(
    y,
    lsl.Dist(tfd.Normal, loc=loc, scale=scale),
    name="y"
)

model = lsl.Model([yvar])
```

image

image

## Minimal example notebook

[test_sample.ipynb.zip](https://github.com/user-attachments/files/19523106/test_sample.ipynb.zip)


tests/model/test_var.py
```diff
@@ -766,3 +767,102 @@ def test_transform_default(self) -> None:
...
+        assert pred.shape[-1] != x.shape[-1]
+        assert pred.shape[-1] == xnew.shape[-1]
+
+
+class TestVarSample:
+    @pytest.mark.parametrize("build_model", [True, False])
+    def test_sample_with_model(self, build_model) -> None:
+        X = lnodes.Var(
+            value=tfp.distributions.Uniform(low=-1.0, high=1.0).sample(
+                (100, 2), jax.random.key(3)
```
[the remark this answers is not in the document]
jobrachem: `jax.random.key` is the current API, `PRNGKey` is legacy: https://docs.jax.dev/en/latest/_autosummary/jax.random.key.html

jobrachem:
## More Elaborate Illustration Notebook (P-Spline Model)

[test_gam.ipynb.zip](https://github.com/user-attachments/files/19523143/test_gam.ipynb.zip)

jobrachem:
@wiep @GianmarcoCallegher since #246 is merged, this PR is now ready for review. Can I request a review from both of you, since I consider this a major piece of functionality?

GianmarcoCallegher:
> @wiep @GianmarcoCallegher since #246 is merged, this PR is now ready for review. Can I request a review from both of you, since I consider this a major piece of functionality?

Of course

liesel/model/model.py
```diff
@@ -1407,6 +1407,186 @@ def simulate(self, seed: jax.random.KeyArray, skip: Iterable[str] = ()) -> Model
...
+            k: v for k, v in posterior_samples.items() if k in vars_and_nodes
+        }
+        try:
+            drawn_samples = draw_samples(filtered_samples, seeds)
+            self.state = state_before
+        except Exception as e:
+            msg = (
+                "Error during sampling. Make sure to check sample shapes! The values in"
+                " 'posterior_samples' must have two leading batching dimensions."
+            )
```
GianmarcoCallegher, line 1586: This error message seems very specific. Can't something else go wrong besides the sample shapes?

liesel/model/model.py
```diff
@@ -1742,3 +1934,68 @@ def load_model(file: str | IO[bytes]) -> Any:
         model = dill.load(file)
 
     return model
+
+
+class TemporaryModel:
+    def __init__(self, *vars_and_nodes, verbose: bool = False, silent: bool = False):
```
GianmarcoCallegher, line 1940: Can we give a little description of this class or make it "private"?

liesel/model/model.py
```diff
@@ -1407,6 +1407,186 @@ def simulate(self, seed: jax.random.KeyArray, skip: Iterable[str] = ()) -> Model
...
+            self.state = self.update_state(position, state_for_sampling)
+
+            sampled_position = {}
+            for name, spec in sampling_specs.items():
+                tfp_dist = spec["dist"].init_dist()
+                value = tfp_dist.sample(spec["shape"], seeds[spec["i"]])
+                sampled_position[name] = value
+                spec["value_var"].value = value
+
+            return sampled_position
```
GianmarcoCallegher, line 1559: It seems counter intutive to me that the function `one_draw`, that is supposed to produce one draw, relies on the parent-enviroment and has a side effect on `self.state`. What did I get wrong?

GianmarcoCallegher (changes requested):
Dear @jobrachem . Thanks a lot for this amazing PR. It looks great. I just left a couple of minor comments

liesel/model/model.py
jobrachem: It is indeed a very specific message. However, the next line is

```python
raise RuntimeError(msg) from e
```

This means, the original error is accessible. The goal is here to add another message, because I think the most likely thing that can go wrong is the wrong shapes.

However, always changing the error to `RuntimeError` might be a bit over the top. I pushed a commit to change this to re-raising the same error class as the original exception.

liesel/model/model.py
jobrachem: Fair point, I added a description. 

Additionally, I would say the class is semi-private, because it is not available as `lsl.TemporaryModel`, so it is not part of the end-user API. Users would have to specifically import `from liesel.model.model import TemporaryModel`.

liesel/model/model.py
jobrachem: The `TemporaryModel` is used only in the `Var.sample` method if the variable is not already part of a model.

But of course: fair point, the side effects are a little sus, and in general, the .sample methods are not jittable at the moment.

Apart from the obvious side-effect when overwrite `self.state`, there's also a side-effect on the model state in the line

```python
spec["value_var"].value = value
```

Both side-effects are, at the moment, important for the function to work as intended. It should be possible to write a version of the function that does not rely on side-effects, but right now I don't have the time to fully think through what would be necessary for that.

liesel/model/model.py
GianmarcoCallegher: I know, but unfortunately, we don't have different level of visibility in Python

liesel/model/model.py
GianmarcoCallegher: Sorry, I overlooked the ‘ TemporaryModel‘ thing. I agree with you that we can keep things unchanged. I just thought that maybe we want to keep this issue in mind for the future ❤️

jobrachem: No worries! :) I totally agree that we probably want to keep the side-effect issue in mind for the future!

GianmarcoCallegher (approved):

liesel/model/model.py
```diff
@@ -1407,6 +1407,247 @@ def simulate(self, seed: jax.random.KeyArray, skip: Iterable[str] = ()) -> Model
...
+
+            # return reshaped version of samples
+            return jax.tree.map(reshape, drawn_samples)
+
+        # this branch of the function continues only if posterior_samples is not None
+        # -----------------------------------------------------------------------------
+        draw_iter = jax.vmap(one_draw, in_axes=(0, 0), out_axes=0)
+        draw_chains = jax.vmap(draw_iter, in_axes=(0, 0), out_axes=0)
+        draw_samples = jax.vmap(draw_chains, in_axes=(None, 0), out_axes=0)
+
```
wiep, line 1623: do you need to restore the state here? otherwise there should be tracers in the state.

liesel/model/model.py
```diff
@@ -1407,6 +1407,247 @@ def simulate(self, seed: jax.random.KeyArray, skip: Iterable[str] = ()) -> Model
...
+        # filter samples to include only samples that belong to the model
+        vars_and_nodes = list(self.vars) + list(self.nodes)
+        filtered_samples = {
+            k: v for k, v in posterior_samples.items() if k in vars_and_nodes
+        }
+
+        try:
+            drawn_samples = draw_samples(filtered_samples, seeds)
+            self.state = state_before
```
wiep, line 1632: this should happen in a finally block so that the state is always restored.

wiep (changes requested):
See two small comments.

liesel/model/model.py
```diff
@@ -1407,6 +1407,247 @@ def simulate(self, seed: jax.random.KeyArray, skip: Iterable[str] = ()) -> Model
 
         return self
 
+    def sample(
+        self,
+        shape: Sequence[int],
+        seed: jax.random.KeyArray,
```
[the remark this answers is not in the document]
jobrachem: Thanks! I updated the type hints and the corresponding docstrings.

liesel/model/model.py
jobrachem: I'm not sure if I understand you correctly. Are you proposing to do the following?

```python
draw_iter = jax.vmap(one_draw, in_axes=(0, 0), out_axes=0)
self.state = state_before
draw_chains = jax.vmap(draw_iter, in_axes=(0, 0), out_axes=0)
self.state = state_before
draw_samples = jax.vmap(draw_chains, in_axes=(None, 0), out_axes=0)
self.state = state_before
```

Or once?

```python
draw_iter = jax.vmap(one_draw, in_axes=(0, 0), out_axes=0)
draw_chains = jax.vmap(draw_iter, in_axes=(0, 0), out_axes=0)
draw_samples = jax.vmap(draw_chains, in_axes=(None, 0), out_axes=0)
self.state = state_before
```

What would be the downside of not doing it? After all, the state gets updated in the `one_draw` helper function anyway.

liesel/model/model.py
jobrachem: You are absolutely right, I pushed a corresponding update.

wiep: actually, this is a compiled function (via vmap), it does will not change the state.

liesel/model/model.py
wiep: I am not sure if that is important here, but if I understand correctly,
one_draw changes the state and does not restore it. during tracing (vmap) the state is changed. now that I give it another thought, it would probably be correct to restore the state in one_draw. Similar to

```
def one_draw(...):
    old_state = model.state
     ....
     model.state = old_state
     return result
```

Like this, the state is restored after tracing. Once tracing is completed, the functions are side-effect free.

liesel/model/model.py
jobrachem: Here's what I observe:

1. Jitting fails. I interpret this as "the function cannot be jit compiled". See the following test:

```python
@pytest.mark.xfail
def test_sample_prior_linreg_jit(self, linreg: Model):
    model = linreg

    # sample with y fixed; i.e. y will not be sampled
    jitted_sample = jax.jit(
        model.sample,
        static_argnames=["fixed", "newdata", "posterior_samples", "dists"],
    )

    jitted_sample(shape=(1, 100), seed=rnd.key(1))
```

2. If the lines `self.state = state_before` are commented out, the model state after function execution is full of jax tracers. I suppose this is what you are referring to here: https://github.com/liesel-devs/liesel/pull/248#discussion_r2067512613. That would suggest that the `vmap` calls have an effect on the state, but the actual calls `draw_cahins()` / `draw_samples()` do not. Still, due to the effect of `vmap`, the state needs to be restored.

liesel/model/model.py
wiep: Leaking tracers can happen, when you trace a computation (jit, vmap, grad, etc) in a function that has side effects. See this example,

```python
import jax

class Model:
  
  def __init__(self):
    self._cache = None

  def do(self, x):
    self._cache = x  # storing a traced value -> tracer leak
    return x
  
  def do_with_cleanup(self, x):
    old_cache = self._cache
    try:
        self._cache = x  # storing a traced value
        return x
    finally:
      print("cleanup")
      self._cache = old_cache  # cleanup the cache -> no tracer leak
  
jax.tree_util.register_dataclass(Model, data_fields=[], meta_fields=[])

model = Model()
x = jax.numpy.arange(5)

# all good
with jax.check_tracer_leaks():
  jax.vmap(model.do_with_cleanup)(x)

# raises exception
with jax.check_tracer_leaks():
  jax.vmap(model.do)(x)
```

The second call to `do` will raise an exception as the tracer gets leaked, i.e., is still referenced when the vmap has returned. To avoid it, the `do_with_cleanup` methods ensures that the cache cannot leak a tracer. If you would jit the function, there is no runtime downside cleaning up the the tracers as this is only done once during compile time. Does that make sense?

jobrachem: Thanks @wiep. I see how a tracer leak can happen, and I think that is what I observed here: https://github.com/liesel-devs/liesel/pull/248#discussion_r2069268929

What I don't understand is: why is that a problem here? After all, the leak is caught by restoring the model state before the function returns.

wiep: I guess it’s not really a problem here as the user will never interact with one_draw, just good practice to avoid leaking tracers.

jobrachem: Thanks for helping me understand this issue. So here's my current reasoning: Resetting the state at the end of each one_draw call would be compiled out of the function we're jitted. At the moment however, it is not jitted and jitting fails (see below). So there might be a small overhead from resetting the state with each call to one_draw. Resetting once before return or each time at the end of one_draw result in the same behaviour for users. For these reasons I would tend to reset the state only once, when the function returns. Would you be on board with proceeding like this, @wiep ?

wiep: let me have a look at the jitting issue.

wiep: I fixed the jitting issue. The problem was that a static value was being converted into a JAX array, which introduced a tracer and caused downstream issues during compilation.

Regarding the state: saving and restoring it are reference operations and essentially negligible compared to the computational cost of `one_draw`. Also, since vmap behaves like a compiler, chaining vmap calls won’t repeatedly invoke state restoration. If I understand the code correctly, when the state is restored in `one_draw`, then the clean up would happen once per `model.sample` call -- or, with jit, once per compilation. So it makes no practical difference whether the state is restored inside `one_sample` or outside of it. That would only change if we used a regular Python loop -- but then tracer leakage wouldn't be a concern in the first place. 

I’ve adjusted the code so that `one_draw` handles the cleanup. One more thing: we might want to consider wrapping the cleanup in a `try ... finally` block, since a failed computation in `one_draw` could otherwise leave the model in an altered state.

If you want, you could test my theory by adding a print stmt and see how often cleanup is printed, .e.g,

```python
def one_draw(...):
     ...
    # to avoid tracer leakage we prevent side effects to persists
    print("cleanup")
    self.state = privous_state
    return sampled_position
```

I think that the following code does 5 cleanups
```python
    @pytest.mark.skip
    def test_sample_prior_linreg_jit_and_resore_state(self, linreg: Model):
        model = linreg

        jitted_sample = jax.jit(
            model.sample,
            static_argnames=["shape", "fixed", "dists"],
        )
        bs = {"b": jnp.linspace(0.0, 1.0, 12).reshape((2, -1, 2))}

        model.sample(shape=(1, 100), seed=rnd.key(1))
        model.sample(shape=(3, 100), seed=rnd.key(1), posterior_samples=bs)
        model.sample(shape=(3, 100), seed=rnd.key(1), posterior_samples=bs)

        jitted_sample(shape=(1, 100), seed=rnd.key(1))

        xx = jitted_sample(shape=(3, 100), seed=rnd.key(1), posterior_samples=bs)["y"]
        xx = jitted_sample(shape=(3, 100), seed=rnd.key(1), posterior_samples=bs)["y"]
```

liesel/model/model.py
wiep: I resolve this in favor of the thread above.

wiep (approved):
@jobrachem, @GianmarcoCallegher, you're happy with the changes I made, we can merge this PR.

liesel/model/model.py
jobrachem: You are correct, your proposed test code prints 5 cleanups. The last line does not trigger new compilation and so it does not print a cleanup.

GianmarcoCallegher:
> @jobrachem, @GianmarcoCallegher, you're happy with the changes I made, we can merge this PR.

We are. Thanks a lot

PR #251, opened 2025-03-30
Add 'name' argument to Var.transform

jobrachem:
With this PR, users can supply a `name` argument to `Var.transform()` in order to provide a custom name to their transformed variable.


jobrachem:
@GianmarcoCallegher I resolved the merge conflicts that arose in the meantime :)

PR #252, opened 2025-04-02
Allow samples dictionary in plotting functions

jobrachem:
The PR does what it says. In addition to

```python
results = engine.get_results()
samples = results.get_posterior_samples()

gs.plot_trace(results)
```

it is now possible to do 

```python
gs.plot_trace(samples)
```

for all plotting functions.

This is really useful in combination with predictions (#246), because you can look at diagnostic plots for computed quantities without having to include them in `eb.positions_included` during sampling, i.e. without having to have them baked into the `gs.SamplingResults` object.

### DIFF
--- a/liesel/goose/summary_viz.py
+++ b/liesel/goose/summary_viz.py
@@ -14,6 +14,8 @@ import seaborn as sns
 
 from liesel.goose.engine import SamplingResults
 
+from .types import Array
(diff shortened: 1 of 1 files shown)


GianmarcoCallegher (approved):
Nice job. Thanks a lot

PR #253, opened 2025-04-04
Future-proof MCMC convenience

jobrachem:
Turns out, I also really like this way of doing it. Things now revolve around the following arguments and objects:

- `gs.MCMCSpec(kernel, kernel_kwargs, kernel_group, jitter_dist)` holds information about MCMC inference for a single variable, but allows for passing a `kernel_group` argument that is used to identify kernels that cover multiple variables.
- `gs.LieselMCMC(model: lsl.Model)` collects MCMC inference information from variables in the model
- `lsl.Var(inference)` new argument that accepts instances or dicts of `MCMCSpec`.

Here's a code example that showcases lots of functionality already:

```python
mu = lsl.Var.new_param(
    0.0,
    lsl.Dist(tfd.Normal, loc=0.0, scale=1.0),
    inference=gs.MCMCSpec(
        gs.NUTSKernel,
        kernel_group="a",
        jitter_dist=tfd.Uniform(low=-1.0, high=1.0),
    ),
    name="mu",
)

sigma = lsl.Var.new_param(
    1.0,
    lsl.Dist(tfd.InverseGamma, concentration=1.0, scale=0.5),
    inference=gs.MCMCSpec(
        gs.NUTSKernel,
        kernel_group="a",
        jitter_dist=tfd.Uniform(low=0.0, high=1.0),
    ),
    name="sigma",
)

model = lsl.Model([mu, sigma])

mcmc = gs.LieselMCMC(model)

mcmc.kernel_list() # list of mcmc kernels
mcmc.jitter_functions() # dictionary of jitter functions
mcmc.engine_builder( # shortcut for an engine builder
    seed=1,
    num_chains=4,
    warmup_duration=200,
    posterior_duration=100,
    term_duration=50, # from here on, the arguments have defaults
    thinning_warmup=1,
    thinning_posterior=1,
    apply_jitter=True
) 
mcmc.engine() # like .engine_builder(), but directly builds the engine
```

From here, it was quite easy to allow multiple specs:

```python
mu = lsl.Var.new_param(
    0.0,
    lsl.Dist(tfd.Normal, loc=0.0, scale=1.0),
    inference={
        "a": gs.MCMCSpec(gs.NUTSKernel),
        "b": gs.MCMCSpec(gs.IWLSKernel),
    },
    name="mu",
)

sigma = lsl.Var.new_param(
    1.0,
    lsl.Dist(tfd.InverseGamma, concentration=2.0, scale=1.0),
    inference=gs.MCMCSpec(gs.IWLSKernel),
    name="sigma",
)

# will take spec 'a' from mu
# and the single spec from sigma
engine = gs.LieselMCMC(model, which="a").engine()
```

### DIFF
--- a/liesel/goose/builder.py
+++ b/liesel/goose/builder.py
@@ -11,23 +11,19 @@ from __future__ import annotations
 
 import logging
 import math
-from collections.abc import Callable, Iterable
-from dataclasses import dataclass, field
-from typing import TYPE_CHECKING, Any, cast
+from collections.abc import Iterable
+from typing import cast
 
 import jax
 import jax.numpy as jnp
-import tensorflow_probability.substrates.jax.distributions as tfd
 
 from liesel.option import Option
 
 from .engine import Engine
 from .epoch import EpochConfig, EpochManager
-from .interface import LieselInterface
 from .kernel_sequence import KernelSequence
 from .pytree import stack_leaves
 from .types import (
-    Array,
     JitterFunctions,
     Kernel,
     KeyArray,
@@ -36,10 +32,6 @@ from .types import (
     QuantityGenerator,
 )
 from .warmup import stan_epochs
-
-if TYPE_CHECKING:
-    from liesel.model import Model, Var
-
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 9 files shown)


jobrachem:
## Behavior during transformation

```python
sigma = lsl.Var.new_param(
    1.0,
    lsl.Dist(tfd.InverseGamma, concentration=2.0, scale=1.0),
    inference=gs.MCMCSpec(gs.IWLSKernel),
    name="sigma",
)

log_sigma = sigma.transform(tfb.Exp(), inference=None) # default: No inference info
log_sigma = sigma.transform(tfb.Exp(), inference=gs.MCMCSpec(gs.NUTSKernel)) # manual
log_sigma = sigma.transform(tfb.Exp(), inference="transfer") # transfer from sigma

sigma.inference # is always None after transformation
```

jobrachem:
@GianmarcoCallegher @wiep 
I drafted a 'maximalist' implementation of the more abstract functionality that we talked about yesterday here. Turns out, I really like this way of setting things up. Thanks for the discussion yesterday! Please let me know your thoughts about this draft when you find the time. Otherwise, of course we can continue the discussion in the next weekly.

jobrachem:
- `Var.transform(inference="drop")` for dropping explicitly; `Var.transform(inference=None)` raises error if there is `Var.inference`.

- Check what happens when `inference="transfer"`, but `Var.inference is None`. Should be an error. 

```
a = Var(...)

inference = a.inference
a.inference = None
b = a.transform()
    a.inference.allow_transformation
b.inference = inference
```


```
a.transform(inference="drop")
```

wiep:
Thoughts from the meeting. Transforming the inference part of `a.transform` could be handled by the object in `a.inference` as this object may have access to additional information about the inference scheme.

jobrachem:
Updated behavior as discussed; described by the three tests below. @wiep @GianmarcoCallegher this is ready for review now :)

```python
def test_transform_var_with_inference_new(self):
    """
    It is allowed to pass a new inferece object during transformation.
    In this case, the inference object of the original variable is removed.
    """
    sigma = lsl.Var.new_param(
        1.0,
        lsl.Dist(tfd.InverseGamma, concentration=2.0, scale=1.0),
        inference=gs.MCMCSpec(gs.IWLSKernel),
        name="sigma",
    )
    inference = sigma.inference

    log_sigma = sigma.transform(tfb.Exp(), inference=gs.MCMCSpec(gs.NUTSKernel))
    assert log_sigma.inference is not inference
    assert sigma.inference is None
    assert log_sigma.inference.kernel is gs.NUTSKernel

def test_transform_var_with_inference_none(self):
    """
    Default behavior when trying to transform a variable *with* inference 
    information: Error. You need to declare explicitly, what you want to do.
    In this case, ``"drop"`` means the inference information is deleted.
    """
    sigma = lsl.Var.new_param(
        1.0,
        lsl.Dist(tfd.InverseGamma, concentration=2.0, scale=1.0),
        inference=gs.MCMCSpec(gs.IWLSKernel),
        name="sigma",
    )
    inference = sigma.inference

    with pytest.raises(ValueError):
        sigma.transform(tfb.Exp())

    log_sigma = sigma.transform(tfb.Exp(), inference="drop")
    assert log_sigma.inference is not inference
    assert sigma.inference is None
    assert log_sigma.inference is None

def test_transform_var_without_inference(self):
    """
    Default when the original variable has no inference information: 
    Everything works smoothly.
    """
    sigma = lsl.Var.new_param(
        1.0,
        lsl.Dist(tfd.InverseGamma, concentration=2.0, scale=1.0),
        name="sigma",
    )
    
    log_sigma = sigma.transform(tfb.Exp())
    
    assert log_sigma.inference is None
    assert sigma.inference is None
    assert log_sigma.inference is None
```

jobrachem:
@wiep @GianmarcoCallegher I updated the PR by removing the duration-arguments from the `gs.LieselMCMC.engine_builder()` method. This way, the `.engine_builder()` method does not include an assumption about the structure of the warmup scheme, and users are nudged to set it themselves with `EngineBuilder.set_duration()` or `EngineBuilder.set_epochs()`.

For the same reason I removed `gs.LieselMCMC.engine()`, because returning a built engine requires assumptions about the structure of the warmup.

liesel/goose/builder.py
```diff
@@ -485,3 +495,131 @@ def build(self) -> Engine:
...
+                )
+
+        return kernel_groups
+
+    def kernel_list(self) -> list[Kernel]:
+        kernel_groups = self.kernel_groups()
+        kernel_list = [
+            g.kernel(g.position_keys, **g.kwargs)  # type: ignore
+            for g in kernel_groups.values()
+        ]
```
GianmarcoCallegher, line 567: Can it be a property or ate least I'd name it `get_kernel_list`

GianmarcoCallegher (commented):
This is great, @jobrachem . I left two small comments

liesel/goose/builder.py
jobrachem: Fair point, thanks @GianmarcoCallegher ! I updated all method names on the `gs.LieselMCMC` object. I prefer `get_` methods here to avoid users to think that these are persistent lists that they could append stuff to.

image

liesel/goose/builder.py
```diff
@@ -485,3 +495,131 @@ def build(self) -> Engine:
...
+        if not isinstance(inference, MCMCSpec):
+            raise ValueError(
+                f"Attribute 'inference' of variable {var} is of type"
+                f" {type(inference)}, but expected type '{MCMCSpec}'."
+            )
+        return inference
+
+    def kernel_groups(self) -> dict[str, _KernelGroup]:
+        vars_ = self.model.vars
+        kernel_groups: dict[str, _KernelGroup] = {}
```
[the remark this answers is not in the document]
jobrachem: Fair point, thanks @GianmarcoCallegher ! I updated all method names on the `gs.LieselMCMC` object. I prefer `get_` methods here to avoid users to think that these are persistent lists that they could append stuff to.

image

liesel/goose/builder.py
```diff
@@ -7,21 +7,27 @@
 implementations.
 """
 
+from __future__ import annotations
```
wiep, line 10: Postponed evaluation of annotations is default since 3.10 (imho). So, we probably don't need it.

liesel/goose/builder.py
```diff
@@ -485,3 +495,131 @@ def build(self) -> Engine:
             quantity_generators=self.quantity_generators,
             show_progress=self.show_progress,
         )
+
+
+@dataclass
+class LieselMCMC:
```
wiep, line 501: just wondering whether this code should go in a new module to keep files a bit smaller. what do you think?

liesel/model/nodes.py
```diff
@@ -32,8 +32,11 @@
 from .viz import plot_nodes, plot_vars
 
 if TYPE_CHECKING:
+    from ..goose import MCMCSpec
     from .model import Model
 
+    InferenceTypes = Union[MCMCSpec, None, dict[str, MCMCSpec]]
```
wiep, line 38: not allowing `any` here, is a bit restrictive if a user has their own inference implementation and wants to use this field. I don't see this as very important right now. However, we could also consider making the types generic over the inference type. Using PEP 695 (v.3.12+) notation that would look like

```python

type InferenceTypes[T] = T | dict[str, T] | None

class Var[InferenceT]:
    def __init__(self, inference: InferenceTypes[InferenceT] = None):
        self->inference = inference
```

Pro: future proof
Neg: more complicated / support of generics is not as nice in < 3.12.

wiep (commented):
Not much to add after our discussion. I like the idea and implementation. Let's briefly consider to be generic over the inference type.

liesel/goose/builder.py
jobrachem: flake8 fails in the pre-commit if I remove the line, so for now I would keep it.

liesel/goose/builder.py
jobrachem: Sure, I moved the new goose code to `goose/mcmc_spec.py`

liesel/model/nodes.py
jobrachem: The PEP 695 (v.3.12+) notation is pretty cool! I did not know about that before. I would like to have that at some point. For now, I'd say let's go forward as it is, but open an issue, ok? I added `Any` to the `InferenceTypes`.

PR #258, opened 2025-04-16
Switch from PyGraphviz to pydot

hriebl:
[Pydot](https://github.com/pydot/pydot) is an alternative to PyGraphviz that is written in pure Python and hence does not require a compiler or the Graphviz headers to install. It only depends on the Graphviz binaries, which makes it a lot easier to use, especially on Windows. As far as I have tested the library, the generated graphs look the same. Combined with https://github.com/liesel-devs/rliesel/pull/9, this PR makes Liesel and RLiesel much more accessible to less tech-savvy people.

Also, I tried to improve the fallback graph layout that is used if Graphviz is missing. Which one do you like better, @GianmarcoCallegher, @jobrachem?

**forceatlas2_layout()**

``` py
node_size = {node: 2.0 for node in graph}
pos = nx.forceatlas2_layout(graph, node_size=node_size, seed=42)
```

forceatlas2_layout

**kamada_kawai_layout()**

``` py
pos = nx.kamada_kawai_layout(graph.to_undirected())
```

kamada_kawai_layout

Both are pretty readable, but the second one doesn't depend on a random seed and might be more stable, I think...

Best,
Hannes

### DIFF
--- a/liesel/model/viz.py
+++ b/liesel/model/viz.py
@@ -135,28 +135,19 @@ def _prepare_figure(graph, width, height, prog):
     fig, axis = plt.subplots()
     fig.set_size_inches(width, height)
 
-    if _is_pygraphviz_installed():
-        pos = nx.drawing.nx_agraph.graphviz_layout(graph, prog=prog)
-    else:
+    try:
+        pos = nx.nx_pydot.pydot_layout(graph, prog=prog)
+    except FileNotFoundError:
         logger.warning(
-            "PyGraphviz was not found in the current environment. "
-            "Using fallback graph layout. Consider installing PyGraphviz: "
-            "https://pygraphviz.github.io/documentation/stable/install.html"
+            "Graphviz not found in PATH. Using fallback graph layout. "
+            "Consider installing Graphviz: https://graphviz.org/download"
         )
-        pos = nx.fruchterman_reingold_layout(graph)
+
+        # node_size = {node: 2.0 for node in graph}
+        # pos = nx.forceatlas2_layout(graph, node_size=node_size, seed=42)
+        pos = nx.kamada_kawai_layout(graph.to_undirected())
 
     return fig, axis, pos
-
-
-def _is_pygraphviz_installed():
-    """Checks if pygraphviz is installed."""
-
-    try:
-        import pygraphviz  # noqa: F401
-    except ImportError:
-        return False
-    else:
-        return True
 
 
 def _add_nodes_with_distribution_to_plot(graph, axis, pos):
--- a/setup.cfg
+++ b/setup.cfg
@@ -26,12 +26,13 @@ install_requires =
     matplotlib>=3.5
     networkx>=2.6
     numpy>=1.22,!=1.24.0,<2.0
+    optax>=0.1.7
     pandas>=1.4
+    pydot>=3.0
     scipy>=1.8.0
     seaborn>=0.13
     tensorflow-probability>=0.17
     tqdm>=4.62
-    optax>=0.1.7
 
 [options.extras_require]
 dev =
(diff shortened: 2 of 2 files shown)


jobrachem:
@hriebl that sounds great! As for the fallbacks, I think I prefer `forceatlas2_layout()`, because it looks more similar to the "usual" layout, since the response is at the bottom.

Just to be clear: If I understand correctly, users will still have to install Graphviz manually, but then afterwards the installation of `pydot` is less tricky than the installation of `pygraphviz`? On mac, the only real difficulty in this step was to provide pygraphviz with the paths to the Graphviz binaries. Is something like this still necessary?

hriebl:
@jobrachem, okay, we can go with `forceatlas2_layout()` for now if that's the majority vote, and evaluate how stable it is in different scenarios.

Also, yes, users still need to install Graphviz manually to use pydot, but only the binaries and not the headers or a C compiler. For PyGraphviz, the headers and a C compiler need to be available *at install time*. In contrast, pydot can be installed completely independently from Graphviz. It will only complain if the binaries are not in the PATH *at runtime*. For this reason, I've opted to add pydot to the package requirements for Liesel.

Having the headers and a C compiler available is not a big deal on Linux or macOS but a major PITA on Windows. See [the installation instructions for PyGraphviz on Windows](https://pygraphviz.github.io/documentation/stable/install.html#windows). So, in summary, the switch probably doesn't make a big difference for us, but for Windows users, it's a great simplification. They can just install Liesel, and if they are unhappy with the fallback graph layout, grab the Graphviz binaries from https://graphviz.org, install those, and end up with much prettier graphs without doing anything else...

GianmarcoCallegher:
I agree with @jobrachem : `forceatlas2_layout()` is closer to the current layout, but can't we also have `kamada_kawai_layout `?

It looks great, and perhaps for some application, it could be more intuitive

jobrachem:
Sorry, I missed the fact that `forceatlas2_layout()` relies on a seed. If that means the behavior is not really predictable for different models, maybe it is not the best option. I agree that `kamada_kawai_layout()` looks clean and readable, too.

hriebl:
From my limited experience, `forceatlas2_layout()` always seems to look kind of "tree-like" but sometimes produces overlaps between the node labels. Also, the orientation is pretty random, so the tree will sometimes point to the sides or upwards. I do have a slight preference for `kamada_kawai_layout()`, because it seems to produce less overlaps and is fully deterministic.

Also, I do agree with @jobrachem: I would prefer not to add any additional arguments to the plotting functions and hence complicate the function signatures further for something that is only meant to be a fallback anyway.

jobrachem:
Ok, I'm on board with `kamada_kawai_layout()` :)

jobrachem:
@hriebl When I run a simple example, I get an error:

```python
import liesel.model as lsl
tau = lsl.Var.new_param(10.0, name="tau")
model = lsl.Model([tau])
lsl.plot_vars(model)
```

image

image

hriebl:
@jobrachem, hmm, your example does work for me. What does `dot -V` return on your machine if you run it in the same venv?

jobrachem:
@hriebl sure:

```
$ dot -V                          
dot - graphviz version 12.2.1 (20241206.2353)
```

jobrachem:
@hriebl networkx version:

```
$ pip list | grep networkx
networkx                      3.3
```

jobrachem:
@hriebl success! I upgraded networkx to version 3.4.2, now the example works. I guess the solution is to put a lower pin networkx>=3.4.2 in place?

jobrachem (approved):

hriebl:
Thanks for taking a look, @jobrachem! I've added another except statement to the code to catch other exceptions in pydot as well. Your previous error would now result in the fallback graph layout instead of failing completely.

GianmarcoCallegher (approved):

wiep:
@jobrachem, I just noticed that the readme might be out of date. It is still referring to `pygraphviz`.

jobrachem:
Good point @wiep. I'll update it.

PR #259, opened 2025-04-28
Unpin numpy, test against multiple python versions in CI

### LINKED ISSUE #199: Unpin numpy version
See this comment:

https://github.com/liesel-devs/liesel/pull/198#issuecomment-2248741496

The issue in the CI does not originate from scipy, but instead from incompatibility of tensorflow probability to numpy 2.0. At first, I could not reproduce the issue locally, because locally, I installed using the `pmc` specs, and pymc currently pins to numpy<2.

The current solution is to pin numpy in liesel to <2.0, too, until the downstream issue is addressed. There's already an issue in the tensorflow probability repo:

- https://github.com/tensorflow/probability/issues/1814

Once this is addressed, we can unpin numpy, since pymc correctly handles its dependency on numpy itself.

_Originally posted by @jobrachem in https://github.com/liesel-devs/liesel/issues/198#issuecomment-2248741496_

jobrachem:
Closes #199

### DIFF
--- a/.github/README.md
+++ b/.github/README.md
@@ -199,16 +199,16 @@ loc
 kernel_00
 </td>
 <td>
--0.095
+-0.113
 </td>
 <td>
-0.452
+0.458
 </td>
 <td>
--0.823
+-0.865
 </td>
 <td>
--0.100
+-0.118
 </td>
 <td>
 0.664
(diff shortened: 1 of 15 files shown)


wiep (commented):
I am wondering if we can bump the requirement to numpy >= 2.0. I think, numpy 2.0 came with a few breaking changes and by updating now we do not need to worry about them / test the code against <2.0 and >= 2.0.

PR #260, opened 2025-04-29
Ruff

wiep:
Replaces linters and formatters with ruff.

A few code adjustments are also needed to make them pass.


jobrachem (approved):
This is great!

Issue #192, opened 2024-06-12
Activate tutorials action


jobrachem:
> I had to disable the tutorials workflow for the time being. The problem is that, on the GitHub action runner, we get reticulate 1.2x, but we need 1.31 or higher for rliesel.

See https://github.com/liesel-devs/liesel/pull/191#issuecomment-2163508495

jobrachem:
https://github.com/liesel-devs/liesel/pull/266

PR #261, opened 2025-04-29
use uv for project management and hatchling for building

wiep:
An attempt to change the project management to uv and the build system to hatchling. 

This PR is continuing #260.


wiep:
I hope the main parts of moving things to uv hatchling are handled in this PR. The hooks that run on this branch are all working. However, I deactivated the caching but that could be re-enabled ([uv docs](https://docs.astral.sh/uv/guides/integration/github/#caching)). Moreover, we need to see whether the remaining hooks are working.

We should also decide whether we want to include the lock file `uv.lock` in the repo or remove it and then also remove the `--locked` flag in the work flows.

jobrachem:
@wiep : Awesome, let's try to move forward with this quickly. #260 is merged, should we rebase this PR to handle the merge conflicts and declutter the diff?

jobrachem:
Consindering the uv.lock file, here's my current understanding:

1. Including the `uv.lock` file is for development purposes. I.e., when we work on the project, we can do `uv venv && uv pip install -e .` to set up the development environment in a fully reproducible way - should drastically reduce "works on my machine" issues.
2. The CI will also use the lock file. While this makes dependency resolution faster and more stable in the CI, it might mean that we do not catch it when things break due to updated dependencies.
3. Users who do `pip install liesel` are not affected by the `uv.lock` file; dependency resolution proceeds as normal. They could however benefit from the stability of `uv.lock` by cloning the github repo and doing `uv venv && uv pip install -e .`

Is this correct @wiep? 

The issue in 2) could be handled by scheduling a weekly CI run with the ``--no-lock`` flag.

If my understanding is correct and we address the issue in 2), I think using `uv.lock` is helpful.

wiep:
Deactivated the tutorials temporarily in d6177a11f7bfa9d23f3319078445e4391aa70107, as the fails seems to be unrelated to the code in this PR.

wiep:
I think you're spot on.

1. Yes, including uv.lock ensures a reproducible environment for development and CI.
2. Also correct that the CI relying on the lock file could mask upstream dependency issues. I like the idea of scheduling a weekly CI run with `--no-lock` as a good way to catch those. The weekly run could also include the test of all mcmc components and potentially more longer lasting tasks; e.g., we might want to avoid building the tutorials on every commit/PR.
3. And yes, end users installing via` pip install liesel` won’t be affected by `uv.lock` it is not included in the distributed version; but cloning the repo and using `uv` would use the lock file.

So as long as we monitor the freshness of dependencies, I agree committing `uv.lock` is a helpful practice.

wiep:
I am slightly confused with all the README files. I updated the `.github/README.qmd` file. Could you generate the other md files or let me know what I should do to get them synced?

jobrachem:
- I generated the `.github/README.md` file. 
- I also tested the uv setup on my machine, works nicely.
- I also test `uv build`, appears to work, too.

jobrachem:
I also created a scheduled test run to run all tests without locking. From my side, we are good to go with this PR, @wiep

jobrachem (approved):

wiep:
Amazing. Then I will merge this PR and we have to re-activate the tutorials as part of https://github.com/liesel-devs/liesel/issues/192.

Issue #262, opened 2025-04-30
Transformation of weak variables is problematic


jobrachem:
Transformation of weak variables was merged in https://github.com/liesel-devs/liesel/issues/235, but it was based on a conceptual error. I think we should remove this for now.

Consider the following problem description. 

_Originally posted by @jobrachem in https://github.com/liesel-devs/liesel/issues/235#issuecomment-2621782240_

@wiep turns out, I commited a thinko with my usecase. I would want the following behavior:

### Target

```python
import liesel.model as lsl
import jax.numpy as jnp
import tensorflow_probability.substrates.jax.distributions as tfd
import tensorflow_probability.substrates.jax.bijectors as tfb

x = lsl.Var.new_obs(jnp.array([[1.0]]), name="x")
b = lsl.Var.new_param(jnp.ones(1), name="b")
s_transformed = lsl.Var.new_calc(jnp.dot, x, b, name="s_transformed") # actually a preliminary model
y = lsl.Var.new_obs(jnp.zeros(1), lsl.Dist(tfd.Normal, loc=0.0, scale=s_transformed), name="y")

# Apply bijector later
s = lsl.Var.new_calc(jnp.exp, s_transformed, name="s")
y.dist_node["scale"] = s

print(s_transformed.value)
print(s.value)
```

```
[1.]
[2.7182817]
```

![image](https://github.com/user-attachments/assets/c892848b-43d1-4f6a-82a6-1298a6e0c446)

### Implemented behavior

```python
import liesel.model as lsl
import jax.numpy as jnp
import tensorflow_probability.substrates.jax.distributions as tfd
import tensorflow_probability.substrates.jax.bijectors as tfb

# initial model definition
x = lsl.Var.new_obs(jnp.array([[1.0]]), name="x")
b = lsl.Var.new_param(jnp.ones(1), name="b")
s = lsl.Var.new_calc(jnp.dot, x, b, name="s") # actually a preliminary model
y = lsl.Var.new_obs(jnp.zeros(1), lsl.Dist(tfd.Normal, loc=0.0, scale=s), name="y")

# Apply bijector later
log_s = s.transform(tfb.Exp())
print(log_s.value)
print(s.value)
```

```
[0.]
[1.]
```

![image](https://github.com/user-attachments/assets/6de2df7e-39b2-4415-a3cf-f926bca0c381)

### The important difference

In the first version, the transformed variable and its output are

```
s_transformed = x @ b
s = exp(s_transformed)
```

[...]

PR #263, opened 2025-04-30
Remove transformation of weak vars

jobrachem:
See this issue:

- https://github.com/liesel-devs/liesel/issues/262

This PR removes transformation of weak vars, but keeps the corresponding tests, which are updated to expect errors now.

### DIFF
--- a/liesel/model/nodes.py
+++ b/liesel/model/nodes.py
@@ -1661,8 +1661,8 @@ class Var:
                 "To proceed with transformation, the .inference information needs to"
                 "be explicitly removed. You can transform with ``inference='drop'``."
             )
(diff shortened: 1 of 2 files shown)


wiep (approved):
can be merged if we want to prevent this behavior.

jobrachem:
As described in the issue: I guess I am a little confused, and did not take the time this morning to think closely enough about it. Maybe the problem is not really in the implementation, just the specific use case described in the issue is not sensible. Let's wait a bit before merging.

jobrachem:
Closing. I think the issue is with the usecase I had in mind, not necessarily with the functionality.

PR #265, opened 2025-04-30
rename Calc(update_on_init) -> Calc(_update_on_init)

jobrachem:
As the title says. Closes https://github.com/liesel-devs/liesel-internal/issues/310


wiep (approved):
I am surprised that there is no documentation that needs to be updated (or did I miss it?). Otherwise, can be merged.

jobrachem:
Thanks for the review! The docstrings for `Calc` and `Var.new_calc` are updated. Merging.

PR #266, opened 2025-04-30
Activate tutorials action

jobrachem:


jobrachem:
@wiep this is ready for review. When merging, you might remove

```
push:
    branches: [main]
pull_request:
    branches: [main]
```

so that the tutorials run only via schedule. I left it in there right now so that the action would trigger in this PR.

One nice thing is that now, all tutorials are run separately. So when one fails, we know exactly which one.

wiep:
#261 has been merged which created a conflict.

jobrachem:
Merge conflict resolved :)

jobrachem:
Seems like after #261, quarto now again has trouble finding liesel. I'm investigating.

jobrachem:
@wiep we are all set with the tutorials now.

PR #267, opened 2025-04-30
Create update-tutorials.yml

jobrachem:
This action renders the tutorials and creates a pull request to commit any changes. This action is supposed to be run only manually to avoid committing a lot of minimally different versions of images to the repository; inflating its size.


jobrachem:
This is part of the main branch now already.

wiep:
Does that mean #192 can be closed?

PR #269, opened 2025-05-01
Deprecate groups, remove Groups tutorial

jobrachem:
As the title says.

The functionality in https://github.com/liesel-devs/liesel/pull/202 and the pattern of using a `gs.LieselInterface` in Gibbs samplers make groups unnecessary. To avoid confusing users, I think we should remove Groups.


wiep:
Should we discuss this change? I am not too familiar with the group interface anymore. I am also ok with merging it. Your call.

wiep (approved):
Should we discuss this change? I am not too familiar with the group interface anymore. I am also ok with merging it. Your call. Especially, since we only deprecate the feature.

PR #272, opened 2025-05-01
Donate model state and kernel states in engine

wiep:
Jax supports donation of arguments to avoid memory reallocation. This PR enables the use in the engine. There shouldn't be a large speed improvements when the states are small and the number of iterations glued into one compiled call is large (like the default). Nonetheless, a small improvement.

### DIFF
--- a/src/liesel/goose/engine.py
+++ b/src/liesel/goose/engine.py
@@ -371,7 +371,8 @@ class Engine:
                 self._sample_many,
                 in_axes=(0, None, 0, 0),
                 out_axes=(None, 0, 0, 0, 0, 0, 0),
-            )
+            ),
+            donate_argnames=("kernel_states", "model_state"),
         )
 
     @property


jobrachem:
@wiep do you want a review for this?

wiep:
Not required. I don't think there are any issues to expect.

jobrachem:
@wiep the scheduled tutorial tests seem to run into an error that is related to the argument donation

```
 jaxlib.xla_extension.XlaRuntimeError: INVALID_ARGUMENT: Attempt to donate the same buffer twice in Execute() (flattened argument 49, replica 0, partition 0, first use: 30). Toy example for this bug: `f(donate(a), donate(a))`.
``` 

https://github.com/liesel-devs/liesel/actions/runs/14855802112/job/41708784062

Could this be related to this PR?

wiep:
absolutely. I forgot to run the tutorial CI. Tests work. Maybe, I was too fast merging. I can't look into it right now. Should we undo the PR?

PR #279, opened 2025-05-09
Liberalize kwargs equality test for gs.MCMCSpec

jobrachem:
The `gs.LieselMCMC` currently strictly tests for all `kernel_kwargs` within a kernel group to be fully equal. This is implemented via:

```
same_kwargs = group.kwargs == inference.kernel_kwargs
```

This test is flaky. While it compares all elements of the two dictionaries, this comparison is not always well-defined. For example, if a value in the dictionaries is an array, this line will cause an error.

I considered three options:

1. Try to develop a complete, recursive test for equality of two dictionaries that covers many edge cases.
2. Abandon the idea of testing for full equality of values, test for equality of keys instead.
3. Assume that, for each group, only one `gs.MCMCSpec` should define the `kernel_kwargs`.

I think 2. is much more appropriate for the problem than 1. I also think that always enforcing 3. is too restrictive, but allowing 3. is nice; especially in combination with 2. So this PR implements a combination of 2. and a liberal 3. 

#### This is possible now:

spec1 determines the kwargs.

```python
spec1 = gs.MCMCSpec(
    gs.NUTSKernel,
    kernel_group="a",
    kernel_kwargs={"mm_diag": True, "da_target_accept": 0.8},
)

spec2 = gs.MCMCSpec(
    gs.NUTSKernel,
    kernel_group="a",
)

gs.LieselMCMC(model).get_kernel_groups()
```

#### This will raise an error

```python
spec1 = gs.MCMCSpec(
    gs.NUTSKernel,
    kernel_group="a",
    kernel_kwargs={"mm_diag": True, "da_target_accept": 0.8},
)

spec2 = gs.MCMCSpec(
    gs.NUTSKernel,
    kernel_group="a",
    kernel_kwargs={"mm_diag": True, "da_target_accept": 0.8},
)

gs.LieselMCMC(model).get_kernel_groups()
```

#### This is allowed

```python
spec1 = gs.MCMCSpec(
    gs.NUTSKernel,
    kernel_group="a",
    kernel_kwargs={"mm_diag": True, "da_target_accept": 0.8},
)

spec2 = gs.MCMCSpec(
    gs.NUTSKernel,
    kernel_group="a",
    kernel_kwargs={"mm_diag": True, "da_target_accept": 0.8},
)

gs.LieselMCMC(model, kwargs_strategy=gs.KwargsStrategy.COMPARE_KEYS).get_kernel_groups()
```

#### This will not cause an error

This case shows the danger. Only one of the settings will be applied. It is hard to predict which one, and users may never know that a setting has no effect.

However, this danger occurs only if users consciously switch off strict mode.

```python
spec1 = gs.MCMCSpec(
    gs.NUTSKernel,
    kernel_group="a",
    kernel_kwargs={"mm_diag": True, "da_target_accept": 0.8},
)

spec2 = gs.MCMCSpec(
    gs.NUTSKernel,
    kernel_group="a",
    kernel_kwargs={"mm_diag": False, "da_target_accept": 0.5},
)

gs.LieselMCMC(model, kwargs_strategy=gs.KwargsStrategy.COMPARE_KEYS).get_kernel_groups()
```

#### This is not allowed

```python
spec1 = gs.MCMCSpec(
    gs.NUTSKernel,
    kernel_group="a",
    kernel_kwargs={"mm_diag": True, "da_target_accept": 0.8},
)

spec2 = gs.MCMCSpec(
    gs.NUTSKernel,
    kernel_group="a",
    kernel_kwargs={"mm_diag": True},
)
```


jobrachem:
@wiep this is built on top of https://github.com/liesel-devs/liesel/pull/274

wiep:
I wonder if we should consider the following approach to make comparisons easier and more consistent.

If two or more MCMCSpec objects are in the same kernel group, they must have the same `kernel` and `kernel_kwargs`. "Same" means they refer to the exact same object (i.e., same `a is b` or  `id(a) == id(b)`). As a shortcut, users can use `None` instead of repeating the same object. See examples below for what's allowed and what isn't.

I think that helps to avoid confusing situations. For example, if one spec is updated during experimentation but the other isn’t, things can quickly go out of sync. I feel that having different comparison strategies makes things here more complicated than it should be.


### Permitted

```{python}
# provide arguments once and just define groups
spec1 = gs.MCMCSpec(
    gs.NUTSKernel,
    kernel_group="a",
    kernel_kwargs={"mm_diag": True, "da_target_accept": 0.8},
)

spec2 = gs.MCMCSpec(
    kernel_group="a",
)

# Explicitly providing the same object
kernel = gs.NUTSKernel
kwargs = {"mm_diag": True, "da_target_accept": 0.8}

spec1 = gs.MCMCSpec(
    kernel,
    kernel_group="a",
    kernel_kwargs= kwargs,
)

spec2 = gs.MCMCSpec(
    kernel,
    kernel_group="a",
    kernel_kwargs=kwargs,
)


# Use the same kernel class but provide kwargs only once
spec1 = gs.MCMCSpec(
    gs.NUTSKernel,
    kernel_group="a",
    kernel_kwargs= kwargs,
)

spec2 = gs.MCMCSpec(
    gs.NUTSKernel,
    kernel_group="a",
)

# different kernel groups
spec1 = gs.MCMCSpec(
    gs.NUTSKernel,
    kernel_group="a",
    kernel_kwargs={"mm_diag": True, "da_target_accept": 0.8},
)

gs.MCMCSpec(
    gs.NUTSKernel,
    kernel_group="b",
    kernel_kwargs={"mm_diag": True, "da_target_accept": 0.8},
)
```


### Not Permitted

```{python}
# kwargs do not have the same id
# error message should explain this
spec1 = gs.MCMCSpec(
    gs.NUTSKernel,
    kernel_group="a",
    kernel_kwargs={"mm_diag": True, "da_target_accept": 0.8},
)

gs.MCMCSpec(
    gs.NUTSKernel,
    kernel_group="a",
    kernel_kwargs={"mm_diag": True, "da_target_accept": 0.8},
)

# Different Kernel Classes
spec1 = gs.MCMCSpec(
    gs.NUTSKernel,
    kernel_group="a",
    kernel_kwargs={"mm_diag": True, "da_target_accept": 0.8},
)

gs.MCMCSpec(
    gs.HMCKernel,
    kernel_group="a",
)
```

jobrachem:
I think that approach makes sense, and I very much appreciate that it reduces the complexity!

wiep (changes requested):
See the discussion starting here https://github.com/liesel-devs/liesel/pull/279#issuecomment-2869932869.

PR #281, opened 2025-05-12
Fix optim indexing

GianmarcoCallegher:


tests/model/test_var_transform.py
jobrachem, line 1: This doesn't seem to be related to the rest of the PR.

jobrachem (commented):
Looks good, just see one minor comment; and the pre-commit is currently failing.

jobrachem (approved):

PR #290, opened 2025-07-08
First draft error correction tutorial

JanParlesak:


jobrachem:
Hey @JanParlesak , in order for your tutorial to show up in the tutorial overview (https://liesel--290.org.readthedocs.build/en/290/tutorials_overview.html), it still needs to be linked in `docs/source/tutorials_overview.rst` :)

docs/source/tutorials/qmd/07-error-correction.qmd
```diff
@@ -0,0 +1,623 @@
...
+
+$$
+\mathbf{u}_i \sim N_M(\mathbf{0}, \mathbf{\Sigma}_{u,i}).
+$$
+
+The fundamental concept behind Bayesian measurement error correction is
+to treat the true, unknown covariate values $x_i$ as additional latent variables.
+These values are then imputed using MCMC simulations while simultaneously
+estimating all other parameters of the model. To achieve this, a prior distribution
+is assigned to $x_i$. Additional flexibility is achieved by adding hyperpriors with
```
jobrachem, line 45: Quite the deep hierarchy. Does tau^2_mu also receive a prior? Or is that one fixed?

docs/source/tutorials/qmd/07-error-correction.qmd
```diff
@@ -0,0 +1,623 @@
...
+
+$$
+\tilde{x}_i^{(m)} = x_i + u_i^{(m)}, \quad m = 1, \ldots, M.
+$$
+
+The measurement error is assumed to follow a multivariate Gaussian model,
+such that
+
+$$
+\mathbf{u}_i \sim N_M(\mathbf{0}, \mathbf{\Sigma}_{u,i}).
```
jobrachem, line 38: A short comment below, mentioning what the Sigma here looks like would be nice. Is that assumed and kept fixed, or does it also have a prior? Something like that.

docs/source/tutorials/qmd/07-error-correction.qmd
```diff
@@ -0,0 +1,623 @@
...
+
+# Scale the Sigma_me matrices by individual variance factor
+sigma_matrices = []
+for i in range(n):
+    sigma_sq_ui = 1 if i < n // 2 else 2  # First half scaled by 1, second half by 2
+    Sigma_me = create_sigma_me(M, c_u)
+    scaled_matrix = sigma_sq_ui * Sigma_me
+    sigma_matrices.append(scaled_matrix)
+
+    sigma_matrices_array = jnp.array(np.stack(sigma_matrices))
```
jobrachem, line 104: Ich denke dieser Teil könnte vereinfacht werden durch so etwas:

```python
# Covariance matrix template
def get_sigma_me(scale):
    base = np.full((M, M), c_u)
    np.fill_diagonal(base, 1)
    return scale * base

# Covariance matrices and replicates
sigma_scales = np.where(np.arange(n) < n // 2, 1.0, 2.0)
sigma_matrices = [get_sigma_me(scale) for scale in sigma_scales]
```

docs/source/tutorials/qmd/07-error-correction.qmd
```diff
@@ -0,0 +1,623 @@
...
+    Sigma_me = sigma_matrices[i]
+    samples = np.random.multivariate_normal(mean_vector, Sigma_me)
+    replicates.append(samples)
+
+# Generate response variable y
+variances = np.random.choice([0.3, 0.5], size = n)
+y_true = np.sin(x)
+y = np.random.normal(loc = np.sin(x), scale = np.sqrt(variances))
+
+# Convert to JAX arrays
```
jobrachem, line 119: Man könnte auch die ganze Erzeugung in Jax durchführen, dann wäre keine Konvertierung hier mehr notwendig. Das fänd ich ganz nett :)

docs/source/tutorials/qmd/07-error-correction.qmd
```diff
@@ -0,0 +1,623 @@
...
+ax1.legend()
+ax1.grid(True, alpha=0.3)
+```
+## Implementing the Model in Liesel
+
+Now that we have our data we can define our distributions from above in Liesel,
+setting $\tau^2_{\mu} = 1000$ and $a_x = b_x = 0.001$.
+We will initialize $x$ using the mean of our observed measurements. We also need to assign a distribution to the observed
+measurements, where we choose a
+[`multivariate normal distribution`](https://www.tensorflow.org/probability/api_docs/python/tfp/distributions/MultivariateNormalFullCovariance?hl=en). Note that we are building
```
jobrachem, line 147: Mir gefällt die Verlinkung! Vllt noch den Link Text als `tfd.MultivariateNormalFullCovariance` schreiben.

docs/source/tutorials/qmd/07-error-correction.qmd
```diff
@@ -0,0 +1,623 @@
...
+mu_x_prior = lsl.Dist(tfd.Normal, loc = location_mu_x_prior, scale = scale_mu_x_prior)
+
+# Define mu_x as a parameter with the prior distribution
+mu_x = lsl.Var.new_param(0.0, distribution = mu_x_prior, name = "mu_x")
+
+# Define prior distribution for x
+x_prior_dist = lsl.Dist(tfd.Normal, loc = mu_x, scale = tau2_x)
+
+# Estimate x using the mean of replicates and assign a prior distribution
+x_estimated = lsl.Var.new_param(jnp.expand_dims(x_tilde.mean(axis = 1), -1), # initial estimation is the mean of the replicates
```
jobrachem, line 175: Hier würde ich die Zeile anders umbrechen:

```python
x_estimated = lsl.Var.new_param(
    jnp.expand_dims(x_tilde.mean(axis = 1), -1), # initial estimation is the mean of the replicates
    distribution = x_prior_dist,
    name = "x_estimated"
)

```

docs/source/tutorials/qmd/07-error-correction.qmd
```diff
@@ -0,0 +1,623 @@
...
+# Compute spline basis matrix for modeling the scale
+basis_matrix_var_scale = lsl.Var.new_calc(
+    lambda x: splines.basis_matrix(
+        x.squeeze(),
+        knots = knots,
+        order = 3,
+        outer_ok = True),
+    x = x_estimated,
+    name = "basis_matrix_scale"
+)
```
jobrachem, line 243: The basis matrix can be the same for both the location and scale, so no need to use two variables here. That should actually reduce computational burden.

docs/source/tutorials/qmd/07-error-correction.qmd
```diff
@@ -0,0 +1,623 @@
...
+# Define x_tilde (observed replicates) as observed data in the probabilistic model
+x_tilde_var = lsl.Var.new_obs(
+    value = x_tilde,
+    distribution = measurement_dist,
+    name = "x_tilde"
+)
+```
+## The Distributional Regression Model
+
+Following that we will set up our distributional regression model using P-splines. Assuming a Gaussian with $N(\mu_i, \sigma_i^{2})$ for the response, the regression parameters are linked to the model parameters by $\mu_i = \eta^{\mu}(x_i)$ and $\sigma_i^{2} = exp(\eta^{\sigma^{2}}(x_i))$, where $exp(\cdot)$ ensures positivity of the scale parameter. The predictors are then given by:
```
jobrachem, line 195: In math mode, write `\exp` instead of `exp`. It is an established function name. Usually, established function names are written upright, not in italics.

docs/source/tutorials/qmd/07-error-correction.qmd
```diff
@@ -0,0 +1,623 @@
...
+scale_of_y = lsl.Var.new_calc(
+    pred_fn,
+    beta0 = b0_scale,                       # Intercept for scale function
+    spline_coef = coef_scale,               # Spline coefficients for scale
+    basis_matrix = basis_matrix_var_scale,  # Basis matrix for scale
+    name = "scale_of_y"
+)
+
+# Transform scale_of_y to ensure positivity (exponential transformation)
+scale_of_y_transformed = lsl.Var.new_calc(jnp.exp, scale_of_y, name = "scale_of_y_transformed")
```
jobrachem, line 344: I would turn the naming around here. This is the actual scale, not the transformed scale. The transformed scale is the log scale (the additive predictor).

jobrachem (changes requested):
Nice work @JanParlesak ! I left a few comments. Maybe for wrapping up the tutorial in the end, it would be nice to see a plot of the estimated nonlinear functions with some credible interval and the true functions. 

You can achieve that by using the `EngineBuilder.positions_included` attribute to save the values of the location and scale variables while sampling. Then you can use `gs.Summary` to compute the averages and the highest posterior density intervals.

docs/source/tutorials/md/01-lin-reg_files/figure-commonmark/generate-data-1.png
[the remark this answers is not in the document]
jobrachem: Hi @JanParlesak, you marked this conversation as resolved, but I do not agree. The PR still makes changes to files that belong to tutorial 1, and it should not do so. Would you have another look at that? I see that the pictures look the same, but this is about keeping updates to binary files like images in our repository as minimal as possible, because any such change increases the size of the repository.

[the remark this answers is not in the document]
jobrachem: Some goes for the other images relating to tutorial 1.

PR #296, opened 2025-07-24
implements an option to provide kernels with a logprob based on subsets

wiep:
Draft. Currently, only HMC has the new feature implemented.

todos
- [ ] discuss design
- [ ] implement in all kernels

notes from using this already in a project:

- it would be good to be able to create the new log prob calculator with a negative definition, that is, which vars should be excluded. this would probably require to add notes first to the graph builder and then use a gb method.
- add an option to create an variable so it shows up in model.plot_vars.
- it is easy to forget to add the new calc to the gb.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [project]
 name = "liesel"
-version = "0.4.0-dev"
+dynamic = ["version"]
 description = "A probabilistic programming framework with a focus on semi-parametric regression"
 readme = "README.md"
 authors = [
@@ -35,7 +35,7 @@ dependencies = [
     "pydot>=3.0",
     "scipy>=1.8.0",
     "seaborn>=0.13",
-    "tensorflow-probability>=0.17",
+    "tfp-nightly>=0.26.0.dev20250811",
     "tqdm>=4.62",
 ]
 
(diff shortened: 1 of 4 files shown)


wiep:
@jobrachem, could we discuss the design soonish? I'd like to adjust the other kernels and then merge this.

jobrachem:
Hey @wiep, the implementation looks solid and straight-forward to use. I have some general thoughts:

- Currently, the implementation requires changes to the interface of all kernels, adding an init argument that is not (and probably never will be) compatible with other interfaces than the LieselInterface.
- I am wondering whether the problem could maybe also be solved on the level of the model interface instead. For the kernels, it would then look like any other interface, no change required. To allow different kernels to use different interfaces would require some changes to the builder - but maybe that would work out. We could, for example, let the `Interface.log_prob` method accept ``**kwargs``, and then allow users to pass such kwargs in `EngineBuilder.add_kernel`. Then, if the `LieselInterface.log_prob` receives a `log_prob_node_name` kwarg, it returns `model_state[log_prob_node_name].value`. Such a design would not require changes to any kernels and open up similar (and more) modifications also for other interfaces. Example:

```python
eb.add_kernel(kernel, log_prob_kwargs={"log_prob_node_name": "my_subset_logp"})
```

What do you think?

PR #308, opened 2025-08-25
Revert model copy

jobrachem:
After having merged https://github.com/liesel-devs/liesel/pull/306, I have now encountered irregular tracer leaks into the model state when using `Model.predict` inside jitted functions. For now, I think it might be best to revert the changes in https://github.com/liesel-devs/liesel/pull/306 until we better understand what is going wrong.

### DIFF
--- a/src/liesel/model/model.py
+++ b/src/liesel/model/model.py
@@ -819,9 +819,7 @@ class Model:
         self._to_float32 = to_float32
         if grow:
             model = (
-                GraphBuilder(to_float32=to_float32)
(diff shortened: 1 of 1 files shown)


GianmarcoCallegher:
> @GianmarcoCallegher, in case a review is needed could you do that?

Sure

jobrachem:
That would indeed be appreciated. Thank you @GianmarcoCallegher ! Should be a very short one.

GianmarcoCallegher (approved):

PR #310, opened 2025-09-10
VI Dimension Validation; prohibit batched distributions

marweda:
The PR request refers to the issue https://github.com/liesel-devs/liesel-internal/issues/362

In short:
The issue lies in the OptimizerBuilder class where the _validate_dimensionality method is currently just a placeholder and does not enforce any checks. This causes a risk of mismatches between the cumulative latent variable dimensions and the event shape of variational distributions when calling add_variational_dist. 

To address this, the method is being designed to validate whether latent dimensions align correctly with a distribution’s event shape, detect cases where users mistakenly rely on batch shape instead, and provide informative error messages guiding them toward proper usage in Liesel VI. 

Proposed solutions include raising descriptive errors for mismatches, suggesting either separate calls for independent variables or the use of multivariate distributions, and ensuring validation happens after the distribution is fully constructed so its batch shape can be accessed.

### DIFF
--- a/tests/experimental/vi/test_builder.py
+++ b/tests/experimental/vi/test_builder.py
@@ -5,15 +5,10 @@ from unittest.mock import Mock
 import jax.numpy as jnp
 import optax
 import pytest
-import tensorflow_probability.substrates.jax.distributions as tfd
 import tensorflow_probability.substrates.jax.bijectors as tfb
+import tensorflow_probability.substrates.jax.distributions as tfd
 
 from liesel.experimental.vi import LieselInterface, OptimizerBuilder
-
-
-
-# --- Fixtures ---
-
 
 @pytest.fixture
 def mock_model_interface():
--- a/tests/experimental/vi/test_fixtures_conjugate.py
+++ b/tests/experimental/vi/test_fixtures_conjugate.py
@@ -4,15 +4,17 @@ Each fixture provides data and configuration for a specific conjugate pair.
 """
 
 import jax.numpy as jnp
+import optax
 import pytest
 from tensorflow_probability.substrates import jax as tfp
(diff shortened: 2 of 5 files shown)


mnsharki:
Thanks for this implementation, I really like it! It also looks like it does, what it is supposed to! I had a few minor thoughts while going through it:

- It might be cleaner to convert both `batch_shape_tensor` and `event_shape_tensor` to tuples right at the start of the function, so that later code and error messages don’t need to repeat that logic.

- For the error messages: maybe include a link to the [TFP guide on distribution shapes](https://www.tensorflow.org/probability/examples/Understanding_TensorFlow_Distributions_Shapes). That would give users a clear reference point for what exactly “event shape” vs. “batch shape” means.

- I’m not sure why matrix-valued event shapes are explicitly disallowed. In principle, they should work, but when testing it with a Wishart Distribution the following error appeared:
``` 
 32 def BIJECTOR_NOT_IMPLEMENTED():  # pylint: disable=invalid-name
---> [33](file:///C:/Users/mnsaa/Desktop/liesel/.venv/Lib/site-packages/tensorflow_probability/substrates/jax/internal/parameter_properties.py:33)   raise NotImplementedError('No constraining bijector is implemented for this '
     34                             'parameter.')

NotImplementedError: No constraining bijector is implemented for this parameter.
``` 
- The error output looks a little odd because the f-strings are broken across multiple lines. Might be easier to read if those were formatted as single strings.

marweda:
@jobrachem thanks for the information! I'll get to the bottom of this

marweda:
@jobrachem I fixed the tests and resolved the merge conflict. From my perspective this PR is rdy

PR #311, opened 2025-09-16
Make Model.parental_submodel more robust

jobrachem:
- Model.parental_submodel creates copies of the nodes to include in the submodel. This makes it work, because otherwise we would run into problems, since every node is allowed to be part of only one model. 
- Currently, the copying is done manually by calling `copy.deepcopy`. The problem is that this does not really make sure that the model is unset from the relevant nodes. This sometimes lead to problems.
- In this PR, I call Model.copy_nodes_and_vars first, and then select the relevant nodes from these copies. This makes sure that the model is properly unset from all relevant nodes.

### DIFF
--- a/src/liesel/model/model.py
+++ b/src/liesel/model/model.py
@@ -1019,11 +1019,12 @@ class Model:
 
             nodes_to_include.add(node)
 
-        copy_of_nodes_to_include = deepcopy(nodes_to_include)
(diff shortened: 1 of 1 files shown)


jobrachem:
@GianmarcoCallegher would you do me the favor of reviewing this PR? It's a short one.

GianmarcoCallegher (approved):

PR #321, opened 2025-11-07
Optim auto transform params

jobrachem:


src/liesel/model/nodes.py
```diff
@@ -938,6 +939,195 @@ def all_input_nodes(self) -> tuple[Node, ...]:
...
+        """
+        if not self._auto_transform_params:
+            return {}
+
+        try:
+            param_props = self.distribution.parameter_properties()  # type: ignore
+        except (AttributeError, TypeError) as e:
+            raise RuntimeError(
+                f"Distribution {self.distribution.__name__} does not provide a"
+                f"parameter_properties()method. Cannot auto-transform parameters. "
```
jobrachem: just a forgotten space before "method"

src/liesel/model/nodes.py
```diff
@@ -938,6 +939,195 @@ def all_input_nodes(self) -> tuple[Node, ...]:
...
+                return
+
+            if param_var.weak:
+                # Weak variables are already computed/transformed, skip silently
+                return
+
+            bijector = get_bijector(param_name)
+
+            if bijector is not None:
+                transformable[param_name] = (param_var, bijector)
```
jobrachem: If I see it correctly, this helper function mutates the transformable dictionary directly. This is fine and works, but it would be easier to maintain if it would act a bit more like a "pure" function, so that the changing of the transformable dictionary happens on the top level of this method.

jobrachem:
Hey @marweda , I'm in a bit of a rush, so I just leave some quick comments. Later more.

> Thanks for the detailed response, @jobrachem ! I now understand the vision for this feature. The points and remarks regarding my comment were valid and I have thought a lot about them, but I'll take it from where you left off to ensure that this discussion remains focused on the project goal and does not become tangled up in itself. I'm of course still open to answer things about what I commented earlier if wanted.
> 
> > * `Dist.biject_parameters` bijects its parameters, potentially with the default parameter bijectors by calling `Var.biject` with the specific bijectors. The `Dist.biject_parameters` does not concern itself with a potential distribution of its parameters.
> 
> If `Dist.biject_parameters` does not concern itself with a potential distribution of its parameters, then one scenario that has not been covered in the pseudocode might result in a conflict. It is a mixed situation where:
> 
> * a `Var` **is used as a Dist parameter**,
> * the same `Var` **has its own distribution**,
> * and **no explicit bijector** was set on the `Var`, i.e.
> 
> ```
> d = Var.new_param(
>     1.0,
>     distribution=SomeDist,
>     bijector=None
> )
> 
> ig = Dist(
>     tfd.InverseGamma,
>     concentration=d,
>     scale=1.0,
>     bijectors="auto"
> )
> ```
> 
> Under the current proposal, the Dist-level `"auto"` then delegates to the default **parameter bijector** of the outer `Dist` (the ig variable), because the `Var` has no bijector, is a parameter, and is not weak.
> 
> From the user's point of view there might be nothing controversial with this setup. The “unconstrain me” intention behind `"auto"` for MCMC or SVI unconstrained optimization aligns naturally here since Dist has the ability to transform its parameters if they are a parameter and not weak.
> 
> Assuming that the user has modeled correctly, the **parameter bijector** instead of the **default event-space bijector** in the above situation would even be the same (i think) of the Var’s _own_ distribution.
> 
> The presence of a prior distribution on the `Var` signals that this variable is a _latent random variable_, not a hyperparameter. For such variables, the “unconstrain me” intention behind `"auto"` would need the TransformedDistribution treatment as currently done in [`_transform_var_with_bijector_class`](src/liesel/model/nodes.py#L2355-L2444) and See [`_transform_var_with_bijector_instance`](src/liesel/model/nodes.py#L2291-L2351). For this the Dist class would need to know about whether the parameter holds the distribution.
> 

I don't see the issue: I think the `Dist` simply retrieves the right bijector, and then calls `Var.biject()` with this bijector. The `Var.biject` method then handles all of the rest, just like the current `Var.transform`. What am I missing?


> In order to support not having `Var.transform(None)` and to compute `Var.auto_transform` eagerly, we would need to change how `Var.auto_transform` is done or not rely on it for `Var(bijector="auto")`. With "relying on it" I mean to not set `Var.auto_transform=True` as a result of Var(bijector="auto"). 

Just to be clear: My proposal is to leave `Var.transform` as it is right now, unchanged, and introduce a new, similar method `Var.biject` that can do pretty much the same, but frees us from the backwards-compatibility constraint.

marweda:
@jobrachem thanks for the quick response! Then let us maybe just leave it for until friday, could be very well that I am missing something that can be pointed out in a conversation.

marweda:
> Just to be clear: My proposal is to leave Var.transform as it is right now, unchanged, and introduce a new, similar method Var.biject that can do pretty much the same, but frees us from the backwards-compatibility constraint.

I see, thanks, for pointing that out, I misunderstood that.

marweda:
@jobrachem @GianmarcoCallegher I pushed a new version given your ideas and feedback which we can discuss on friday or in the days until then. There is a test_biject_api.ipynb . I haven't checked if the nodes.py pytests run, that still has to be done. Also I will have to take another look at it in the next days, as I think one case is missing, but on another day. I hope that I am now on the same page as you regarding that feature. If not, please let me know. Have a nice weekend!

PR #338, opened 2025-11-27
Update bijection behavior

jobrachem:
This is essentially https://github.com/liesel-devs/liesel/pull/321, but it cherry-picks the bijection-related commits in order to point the PR directly at the main branch instead of the work-in-progress optim branch.

### DIFF
--- a/src/liesel/goose/iwls.py
+++ b/src/liesel/goose/iwls.py
@@ -4,7 +4,7 @@ Iteratively weighted least squares (IWLS) sampler
 
 from collections.abc import Callable, Sequence
 from dataclasses import dataclass, field
-from typing import ClassVar, Self
+from typing import ClassVar, Literal, Self, get_args
 
 import jax
 import jax.numpy as jnp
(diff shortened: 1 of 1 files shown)


src/liesel/model/nodes.py
```diff
@@ -1000,6 +1030,234 @@ def update(self) -> Dist:
...
+            for i, bijector in enumerate(bijectors):
+                if i >= len(param_names):
+                    break
+                param_name = param_names[i]
+                if bijector == "auto":
+                    bijector_dict[param_name] = default_bijectors.get(param_name)
+                else:
+                    bijector_dict[param_name] = bijector
+        else:
+            raise TypeError(f"Invalid bijectors type: {type(bijectors)}.")
```
jobrachem: Nice type handling

jobrachem (commented):
Nice work @marweda! I especially like how thorough you are with the doc strings :) I made some comments and have not had the chance to look at the notebook yet. Things are looking good, I think we mainly have to tune some details and add unit tests :)

PR #343, opened 2025-12-14
Docs improvements

jobrachem:
Some small but nice improvements:

- Better copybutton for doctest codenblocks (the copied code can be directly executed)
- Added intersphinx links to pandas documentation
- Display defaults next to type hints (avoid breaking formatting if line breaks are not included explicitly in docs)


jobrachem:
Further changes:

- Use type Union operator `|`
- Reorganized the docs. This is a major change of the layout of the documentation. I think things are much easier to find now.

PR #359, opened 2026-03-11
Implement Node value conversion

### LINKED ISSUE #344: Should lsl.Var and lsl.Value expect ArrayLike inputs and apply jnp.asarray?
This would be nice for at least two reasons:

- Users could pass a `pd.Series` directly, and it would be automatically converted to a jax array.
- Currently, there are some typing hickups when using 64 bits. In some cases, when you just pass a float as in `lsl.Value(1.0)`, this will not end up with the correct type if it get fed to tensorflow probability before being converted into a jax array. The solution currently is to do `lsl.Value(jnp.array(1.0))` manually, but it would actually be very convenient if that got done under the hood automatically.

I think if we do this, we could probably even get rid of the `to_float32` argument to models. It keeps causing slight headaches in the sense that we often have to think about passing it when initializing models.

My questions is basically, is there any reason why we would not want to do this? Would like to hear your input @wiep and @GianmarcoCallegher

jobrachem:
This is a minimal implementation draft for #344 : 

1. It introduces an init argument `convert` that accepts a function for converting the supplied value. Conversion is applied in the `Value.value` setter and the `Node._to_node` method.
2. The default for the argument is `jnp.asarray` (note: *not*  `jax.tree.map(lambda x: jnp.asarray(x))`)
3. The PR in its current state does not include global configuration for the default. I think this is not that much of a drawback, because a) `jnp.asarray` is very likely the desired conversion function for most usecases, b) if `jax.tree.map(lambda x: jnp.asarray(x))` is desired, it is most likely required only for individual variables and not all variables in a model, c) if a fully different conversion function is desired as the default for a full model, this can already be implemented without a lot of hassle by subclassing `lsl.Var` and overriding the default in the init.
4. Documentation for the conversion function is currently missing.
5. The default for `GraphBuilder(to_float32)` and `Model(to_float32)` was changed to `False`, because this task is achieved by the default of the `convert` argument instead. 
6. In  https://github.com/liesel-devs/liesel/issues/344#issuecomment-3872713468, @wiep suggested that we could use this opportunity to get rid of weakly typed arrays. This PR currently does not include this feature.

Notably, now float64 can be achieved without hassle by simply setting:

```python
jax.config.update("jax_enable_x64", True)
```

Also, now pandas Series and lists can be passed to lsl.Var, and are converted to jax.Array objects automatically without a need to call `series.to_numpy()` and `jnp.asarray(list)`.


wiep:
>  c) if a fully different conversion function is desired as the default for a full model, this can already be implemented without a lot of hassle by subclassing lsl.Var and overriding the default in the init. Documentation for the conversion function is currently missing.

It would probably require monkey patching lsl.Var because other variables with specific meaning are subclasses of lsl.Var. Still, I think we haven't really seen a use case, so I am ok with a opinionated implementation. If a global configuration is required, this can imho be added later without breaking changes.

> In https://github.com/liesel-devs/liesel/issues/344#issuecomment-3872713468, @wiep suggested that we could use this opportunity to get rid of weakly typed arrays. This PR currently does not include this feature.

Is there a reason why this should not be included? a function to convert to strong types would look like

```python
def to_strong_jax(x):
    if isinstance(x, bool):
        dtype = jnp.bool_
    elif isinstance(x, int):
        dtype = jnp.int32         # should this be changed to 64 bit when 64 flag is set?
    elif isinstance(x, float):
        dtype = jnp.float32      # or use jnp.ones().dtype to dynamically respect 64bit flag 
    elif isinstance(x, complex):
        dtype = jnp.complex64 # two float32
    else: 
        return jnp.asarray(x)  # converting an np array to jnp gives strong typed value. needs to be confirmed.
```

Otherwise, good to go I think.

jobrachem:
>It would probably require monkey patching lsl.Var because other variables with specific meaning are subclasses of lsl.Var.

True. Such monkeypatches may be easier if the default conversion function was stored in a class attribute.

>Still, I think we haven't really seen a use case, so I am ok with a opinionated implementation. If a global configuration is required, this can imho be added later without breaking changes.

I agree.

Concerning the weak type conversion: I am mostly not sure that I fully understand what edge cases may occur, so I did not feel confident implementing conversion to strong types for the moment. Jax is intentional about its weak dtypes (https://docs.jax.dev/en/latest/type_promotion.html), as far as I understand they prevent unintended upcasting when working with Python literals. So if we treat all values as strong types, we decide to not make use of this behavior. Would you be ok with opening a dedicated issue about the weak/strong types to focus the conversation there?

wiep:
The issue is that the weak types will not survive as output from a jit compiled function. As the step function outputs the updated state, it removes the weak flags. thus, the next time the function is called, upcasting might happen + a second compilation is triggered. 

The current design already removes the weak typed state and thus triggers upcasting.

the flow looks like

```
state (with weak type)
  -> call jitted function (no upcasting)
  -> updated state (jitted function returns state without weak types)
  -> call jitted function (now with upcasting, since state has no longer weak types)
```

I don't mind moving this discussion into a separate issue: #364

jobrachem:
Thanks @wiep!

I commented in #364.

For the current PR, I moved the default conversion function into a class attribute and added docstrings.

Issue #206, opened 2024-08-28
Should Liesel.model be more strict about consistent parameter and observed nodes?


jobrachem:
At the moment, Liesel lets you initialize a variable with a distribution without declaring it as a parameter or an observed variable. This can lead to inconsistencies in the model, such that it could be that:

```
lsl.Model.log_prob ≠ (lsl.Model.log_lik + lsl.Model.log_prior)
```

i.e. the posterior is not equal to the likelihood plus the prior. This is, because the `log_lik` and `log_prior` attributes rely on the `lsl.Var.parameter` and `lsl.Var.observed` attributes to identify the variables to count for the respective values.

Possible options would be:

- Add `parameter: bool` and `observed: bool` as init arguments to `lsl.Var`. 
    - raise an error with both are true.
    - raise an error if the variable is being initialized with a distribution and both `parameter` and `observed` are false.
    - Loophole: For parameters with a constant prior, you do not usually specify a distribution. So these parameters are not covered by this case.
- Check for consistent decomposition of the `log_prob` in the `lsl.Model` initialization. Raise an error if the decomposition is inconsistent.
    - Loophole: If a Var with `.parameter=False` just so happens to have `Var.log_prob = 0.0`, the test will not discover an inconsistency in the log prob decomposition caused by this var.

jobrachem:
In #360, we added a validation that logs a warning in case of inconsistency.

Issue #316, opened 2025-10-14
Include acceptance probability / rate in gs.Summary


jobrachem:
- Having observed acceptance rates in the Summary would be useful for diagnosis.
- Empirical acceptance rate (share of "position moved") would probably be most robust, but we do not have it for HMC and NUTS, since blackjax does not return them. We would need to compute them post-hoc. Maybe arviz already has something on offer for that: https://python.arviz.org/en/latest/getting_started/WorkingWithInferenceData.html
- Acceptance probability would probably we easier to implement right away, but may not be available or sensible for all kernels.

jobrachem:
Included in #360

PR #360, opened 2026-03-17
Collection of small updates

jobrachem:
In this PR, I collected a few smallish updates:

1. Add `Model.plot()` and `Var.plot()` as aliases of `.plot_vars()`. This is mostly convenience. But it also strengthens the view that most users will focus on the statistical model. Providing `Model.plot()` gives a clear indication that this is the "main" plot, or the one that "you will probably want to look at first".
2. Rename `Var(distribution)` -> `Var(dist)`. Convenience, but also consistent with the naming of `lsl.Dist`. Implemented in a backwards-compatible way.
3. Add `Model.pop_vars()` and `Model.copy_vars()`. In most cases, I only ever find myself using the `vars` returned in the current `.pop_nodes_and_vars()` method, and when I show this to people, I also focus on the `vars`. This also fits into the view that most users will focus on the statistical model and makes explanations a bit easer, because there is no need to say "ignore this part for now" (referring to the nodes).
4. Updated wording in graph plot legend: "Used in value" -> "Deterministic", "Used in distribution" -> "Stochastic". The previous wording struck me as a little too colloquial.
5. Using dotted linetype for stochastic dependence instead of gray color. I think this improves clarity. Also, the legend for the edge case "Deterministic and stochastic" dependence is shown only when such a case is present in the model. This also improves clarity.
6. A warning is logged, if the Model.log_prob cannot be decomposed into Model.log_prior + Model.log_lik. Variables that are maybe not correctly flagged as parameters or obs (i.e. they have a distribution, but no flag is set) are also logged. This is intended to help users create consistent models with correct Model.log_prior and Model.log_lik. The validation can be turned off with a flag in the Model init. See screenshot below. Addresses #206 

Updated graph:
image

Warning on init:

image


jobrachem:
@wiep could you give this a look? I think most of this is uncontroversial. Most likely to be controversial are, I think: 

1. `lsl.Var(distribution)` -> `lsl.Var(dist)` (Point 2)
2. The validation of log prob decomposition in Model init (Point 6).

jobrachem:
P.S.: New changes included in this PR:

1. `Var.new_calc` has received the argument `cache` for easier initialization of uncached calculator variables. Docs:

```
        cache
            If ``False``, this variable will not store a cache of its value. This means,
            the ``function`` is evaluated every single time that the value of this
            variable is requested. This can save memory, if the computations are
            trivial (such as prepending an axis to an array), but it can greatly slow
            down computations otherwise (such as when the function performs a
            matrix inversion). Internally, if ``cache=True``, this variable wraps a
            :class:`.Calc`, and if ``cache=False``, it wraps a :class:`.TransientCalc`.
```


2. `Model.diagnose()` and `Var.diagnose()` return diagnostic dataframes to help debugging a complete model or the submodel implied by a variable, respectively. The dataframes contain one row for each variable and list, for example, the number of NaN and Inf values, the minimum and maximum, and the dtype of the variable's value and log prob. I think this simple method can be very helpful for debugging. Edit: Added a `verbose` argument, which is False by default.

Not verbose:
image


verbose (truncated):
image


4. `Var.all_input_vars` and `Var.all_output_vars` (and the corresponding `_nodes` methods) gain the arguments `to` / `of`, allowing users to select between inputs to the value or the dist node or both. (Same for outputs.) Just a small convenience update for programming with Liesel vars.

5. Both `Model.nodes` and `Model.vars` now use the topological ordering of nodes in the respective graphs. The topological ordering of vars is now depended upon in `gs.LieselMCMC` to define the default order of MCMC kernels in the algorithm as a reverse topological sort. This comes closer to the target of "starting to sample close to the data" as the previous method, which just used the ordering of the (not explicitly/reliably ordered) `Model.vars`.

6. `gs.LieselMCMC.run_for_epochs()` gains the following arguments:

```
        store_kernel_states
            Whether to store kernel states in sampling results, which may be useful
            for debugging.
        show_progress
            Whether to show progress bars during sampling.
        positions_included
            List of additional position keys that should be tracked, see
            :attr:`.EngineBuilder.positions_included`.
        positions_excluded
            List of position keys that should not be tracked. Excluded keys override
            additional keys see :attr:`.EngineBuilder.positions_excluded`.
```

jobrachem:
The PR now also includes the following updates to `plot_vars`:

1. Slightly bigger nodes, so short variable names have a better change of fitting into the colored area.
2. Removed the class name from nodes to declutter.
3. The legend now uses the "o" marker for node color instead of lines.

image

jobrachem:
Added the following to gs.Summary:

1. In the error summary, kernels are now also identified via their position keys (if a kernel governs multiple position keys, the keys are joined with commas).
2. The summary now includes information on the acceptance probabilities (addresses #316 )

Example output: (Note that position_moved is NaN for NUTSKernels and HMCKernels, because they only report the probability, not the actual decision):

image

jobrachem:
The `gs.SamplingResults` class has gained the following methods:

```python
    def get_warmup_acceptance_probabilities(self) -> dict[str, Array]:
        """
        Returns dictionary of acceptance probabilities during warmup by kernel.
        """

    def get_warmup_position_moved(self) -> dict[str, Array]:
        """
        Returns dictionary of transition movements (0: no move, 1: move)
        during warmup by kernel.
        """

    def get_posterior_acceptance_probabilities(self) -> dict[str, Array]:
        """
        Returns dictionary of acceptance probabilities during posterior by kernel.
        """

    def get_posterior_position_moved(self) -> dict[str, Array]:
        """
        Returns dictionary of transition movements (0: no move, 1: move)
        during posterior by kernel.
        """


    def get_warmup_kernel_states(
        self, process_state: Callable[[KernelState], Any] = asdict
    ) -> dict[str, Any]:
        """
        If available, returns a dictionary of kernel states recorded during
        warmup, organized by kernel.

        The argument ``process_state`` is a callable that is used to process the kernel
        states. The default kernel states in Liesel are dataclasses, which is why the
        default here is ``dataclasses.asdict``.
        """
```

jobrachem:
The list continues to grow:

1. `lsl.TemporaryModel` is a bit less chatty now by default (logs a message at level `info` if variables are temporarily named, and at level `debug` if nodes are temporarily named; except if `verbose=True`)
2. `gs.plot_trace` and related plotting functions had side-effects on the model plots by globally changing the seaborn/matplotlib theme. This is fixed.
3. The model can now also be initialized via `lsl.Model(var)` (previously only `lsl.Model([var])` was possible. This is backwards-compatible.
4. The thinning argument in `gs.LieselMCMC.run_for_epochs` was passed along to the enginebuilder incorrently. This is fixed.
5. Added `cache_path` argument to `gs.LieselMCMC.run_for_epochs`, allowing for low-friction caching of sampling results.
6. Added `Model.copy()`

PR #361, opened 2026-03-20
Allow changes to existing model graphs

jobrachem:
This PR allows existing model graphs to be changed. 

It essentially introduces a new decorator `@changes_model_graph` that replaces the existing `@no_model_method` and `@no_model_setter` decorators. This decorator wraps methods and attribute setters and triggers a re-discovery of the model's graph structure after changes. By default, the re-discovery operates eagerly, but it can be made lazy through a flag on the model. In this case, `Model.update_graph()` must be called manually.

Previously forbidden, now allowed abilities include:

- Rename a node or variable that belongs to a model
- Apply `Var.biject` to a variable that belongs to a model
- Change inputs of nodes and variables that belong to a model

New methods include

- `Model.replace()`, which can be used to replace all usages of one variable with another node or variable throughout the model.
- `Model.add(*vars_and_nodes)`, which adds variables/nodes.
- `Model.join(*models, copy: bool = False)`, which adds all variables/nodes from the supplied models to the calling one.
- `Model.update_graph()`, which does the updating.
- `Model.drop_singletons()`, which can be called manually to drop all singleton nodes from the model.

Currently, the old behavior is maintained by the flag `Model.locked`, which defaults to `True` and forbids changes to an existing model just like before.

- [x] Set Model.locked to False by default.

Some background:

I had been working on it at the beginning of the year mostly to see how it goes, after I had a chat with @hriebl about the possibility. I think it turned out quite well, and I think it can be very useful: It removes friction for users who might not understand why it would be necessary to distinguish between a "build mode" and a "compute mode" for the model. So since I have been making a lot of smaller usability and consistency proposals in the last few days, I thought I'd propose to merge this branch as well.

I added a lot of unit tests, and [this notebook](https://github.com/liesel-devs/liesel/blob/mutable-model/tests/model/test_model_modification.ipynb) is a very useful illustration of the model's behavior when it is changed.


jobrachem:
There's also a new method  `Model.rebuild_graph()`, which rebuilds the model graph completely from a) the originally supplied nodes and vars, saved in `Model.seed_nodes_and_vars` or b) nodes and vars passed to the method directly.

GianmarcoCallegher:
Thanks a lot, @jobrachem . I think you did a great job. I have just a couple of comments 😄

GianmarcoCallegher:
When replacing a variable, it would be nice to allow for the same name:

```python
x = lsl.Var.new_obs(jrd.normal(jrd.key(1), (10,)), name="x")
scale = lsl.Var.new_param(1.0, name="scale")

y = lsl.Var.new_obs(
    jrd.normal(jrd.key(2), (10,)), lsl.Dist(tfd.Normal, loc=x, scale=scale), name="y"
)

model = lsl.Model([y])
model.locked = False

x_ = lsl.Var(
    jrd.normal(jrd.key(13), (10,)),
    name="x_"
) # would be nice to allow the same name
model.replace(x, x_)
```

GianmarcoCallegher:
In this case, if I replace `y` with itself, the only variables that survive are `a`, `b` and `scale`, which is a bit weird in my opinion. I think the model should stay unchanged:

```python
x = lsl.Var.new_obs(jrd.normal(jrd.key(1), (10,)), name="x")
scale = lsl.Var.new_param(
    1.0,
    lsl.Dist(
        tfd.InverseGamma,
        concentration=lsl.Var.new_param(1.0, name="a"),
        scale=lsl.Var.new_param(1.0, name="b"),
    ),
    name="scale",
)

y = lsl.Var.new_obs(
    jrd.normal(jrd.key(2), (10,)), lsl.Dist(tfd.Normal, loc=x, scale=scale), name="y"
)

model = lsl.Model([y])
model.locked = False
model.plot_vars()

model.replace(y, y) # I am not sure about this
model.plot_vars()
```

GianmarcoCallegher:
Right now, we can't replace an observed variable with another one with a different shape. I get why, but maybe this is a bit too restrictive. I think we can allow that, and if the shapes are wrong, this would come up when building the model.

GianmarcoCallegher:
Honestly, I don't know how we should behave in the case that the user tries to replace a bijected variable. Changing the parent variable of a weak variable can lead to a lot of problems:

```python

x = lsl.Var.new_obs(jrd.normal(jrd.key(1), (10,)), name="x")
scale = lsl.Var.new_param(
    1.0,
    lsl.Dist(
        tfd.InverseGamma,
        concentration=lsl.Var.new_param(1.0, name="a"),
        scale=lsl.Var.new_param(1.0, name="b"),
    ),
    name="scale",
)
y = lsl.Var.new_obs(
    jrd.normal(jrd.key(2), (10,)), lsl.Dist(tfd.Normal, loc=x, scale=scale), name="y"
)

model = lsl.Model([y])
model.locked = False
model.plot_vars()

scale.biject(tfb.Exp())
model.plot_vars()

model.replace('h(scale)', 0.0) # not the wisest thing to do

model.plot_vars() # we only have 3 variables, scale, x and y, but scale is still regarded as weak
```

GianmarcoCallegher:
Concerning the `join` method, I don't have many comments. I see why it can be useful, and after testing it, I didn't find any main issues. I am just worried that the user might think that the model is joining by name, which is not the case, because this snippet of code throws an error:

```python
x = lsl.Var.new_obs(jrd.normal(jrd.key(1), (10,)), name="x")
scale = lsl.Var.new_param(
    1.0,
    lsl.Dist(
        tfd.InverseGamma,
        concentration=lsl.Var.new_param(1.0, name="a"),
        scale=lsl.Var.new_param(1.0, name="b"),
    ),
    name="scale",
)
y = lsl.Var.new_obs(
    jrd.normal(jrd.key(2), (10,)), lsl.Dist(tfd.Normal, loc=x, scale=scale), name="y"
)

model = lsl.Model([y])
model.locked = False
model.plot_vars()

x_ = lsl.Var.new_obs(jrd.normal(jrd.key(1), (10,)), name="x")
scale_ = lsl.Var.new_param(
    1.0,
    lsl.Dist(
        tfd.InverseGamma,
        concentration=lsl.Var.new_param(1.0, name="a_"),
        scale=lsl.Var.new_param(1.0, name="b_"),
    ),
    name="scale_",
)
y_ = lsl.Var.new_obs(
    jrd.normal(jrd.key(2), (10,)), lsl.Dist(tfd.Normal, loc=x_, scale=scale_), name="y_"
)

model_ = lsl.Model([y_])
model_.locked = False
model_.plot_vars()
```

jobrachem:
Thanks for your review @GianmarcoCallegher ! Thank you so much for providing example code snippets, that made my job in testing additional changes so much easier! Here are my responses:

1. Model.replace() can now handle replacement of old with new var, if both have the same name.
2. Model.replace() can now handle replacement of a var with itself (does nothing)
3. What you describe about the bijected Var is, I think actually the expected behavior, and it affects not only bijected vars but all weak vars. In `model.replace('h(scale)', 0.0)`, the var `h(scale)` gets replaced with a new node that wraps the value `0.0`. So `scale` is actually still weak, but because its input is now a node (not a var), the input does not show up in the graph anymore. But I get that this leads to confusing model graphs, so I changed the behavior of Model.replace() as follows: If a raw value is supplied as replacement, like in `model.replace('h(scale)', 0.0)`, now a *Var* is initialized instead of a *Node*. Also, the newly initialized Var inherits the name of the Var it is replacing.
4. I think there is a typo in your example code, because the name of the `"x"` variable is still the same in both models, which is why there is an error when trying to join them. After changing the name to `"x_"`, `Model.join()` works without a problem.

jobrachem:
Hey @GianmarcoCallegher, after our talk yesterday I now made the following modifications:

1. `Model.join()` acts more like an actual "joining" method. 
    - It now accepts only a single second model.
    - By default, duplicate names are resolved by renaming the offending variables in *both* models.
    - Users can supply a sequence of variable names to join `by`. Variables listed here are taken from the calling model and replace the corresponding variables of the same name in the second model.
    - There's some error handling and logging to prevent mistakes and keep the user informed.
2. New method `Model.join_by_all()` is like `Model.join()`, but automatically joins by all duplicate names.
3. `Model.add()` now accepts not only vars and nodes, but also models. If a Model instance is added, the behavior is pretty much that of the previous `Model.join()`: Everything from the second model is added, and there's an error if there are duplicate names.

Also, we didn't talk about another point you made:

> Right now, we can't replace an observed variable with another one with a different shape. I get why, but maybe this is a bit too restrictive. I think we can allow that, and if the shapes are wrong, this would come up when building the model.

I added a test for this. As far as I can see, the expected behavior you describe is already happening. When `Model.replace()` is called for variables with different shapes, an error occurs only if the differing shapes lead to errors in evaluating the model graph.

GianmarcoCallegher (approved):
Nice job 😄

PR #367, opened 2026-04-15
Update automatic Var and Node naming (fixes #366)

### LINKED ISSUE #366: Duplicate node names in joint model after `pop_nodes_and_vars()`
Building two separate models with shared node name patterns and then combining them into a joint model raises a duplicate node name error. This occurs even after calling `pop_nodes_and_vars()` on each individual model beforehand.

**Steps to Reproduce:**
```python
import liesel.model as lsl
import liesel_gam as gam

df1 = gam.demo_data(100)
df2 = gam.demo_data(100)

tb1 = gam.TermBuilder.from_df(df1)
tb2 = gam.TermBuilder.from_df(df2, prefix_names_by='ff')

eta1 = gam.AdditivePredictor('eta1')
eta1 += tb1.lin('x_lin')

eta2 = gam.AdditivePredictor('eta2')
eta2 += tb2.lin('x_lin')

y1 = lsl.Var.new_obs(
    value=df1['y'].to_numpy(),
    distribution=lsl.Dist(tfd.Normal, loc=eta1, scale=1.0),
    name='y1',
)
y2 = lsl.Var.new_obs(
    value=df2['y'].to_numpy(),
    distribution=lsl.Dist(tfd.Normal, loc=eta2, scale=1.0),
    name='y2',
)

lsl.Model([y1]).pop_nodes_and_vars()
lsl.Model([y2]).pop_nodes_and_vars()

model = lsl.Model([y1, y2])  # raises duplicate node name error
model.plot_vars()
```

**Expected Behavior:**
`pop_nodes_and_vars()` should fully release all nodes from the model, allowing them to be combined into a new joint model without name conflicts.

**Actual Behavior:**
The joint model raises a duplicate node name error.

**Suspected Cause:**
`TransientNode` names appear to be fixed upon model instantiation and are not reset when `pop_nodes_and_vars()` is called, causing conflicts when building the joint model.

**Possible Workaround:**
None found. Using `gam.TermBuilder.from_df(df2, prefix_names_by='ff')` to prefix node names in the second model does not resolve the issue, suggesting the duplicate names originate from TransientNodes that are unaffected by prefixing.

It is possible to iteratively rename duplicate nodes (as identified by the error message) similarly to `gam.term_builder._format_name`, which makes use of `lsl.model.TemporaryModel` - yet an approach closer to the source would be much preferred.

jobrachem:
This PR updates the mechanism by which automatically generated names are created for variables and nodes. It now works as follows.

In `liesel.model.names`, there is a long list of colors and a long list of animal names. An auto-generated name is created with the pattern

```python
f"_{color}-{animal}-{uuid4().hex[:6]}"
```

where the color and animal are randomly selected from the respective list.

- The color-animal combination makes the names somewhat human-readable, which could aid in debugging. 
- The uuid component additionally serves to make name clashes extremely unlikely.
- The leading underscore serves as a soft signal for auto-generation. If users follow this convention and do not manually name any vars/nodes with leading underscores, this can be used to easily identify auto-generated names if necessary.

Examples for auto-generated names:

- `'_platinum-mandrill-4c2d3d'`
- `'_silver-halibut-1ec76d'`
- `'_viridian-moose-9182e3'`

As before, name auto-generation happens during model initialization in `lsl.GraphBuilder._set_missing_names()`. This method now simply calls the new methods `lsl.Var.ensure_name()` and `lsl.Node.ensure_name()` on all vars and nodes.

The method is very simple:

```python
class Var:
    def ensure_name(self) -> Self:
        """
        Ensures that the variable has a name.

        If the variable already has a name, nothing happens. Otherwise, a unique random
        name is generated with a leading underscore.
        """
        if self.name:
            return self
        else:
            self.name = "_" + random_name()
            return self
```


wiep (approved):
looks like a useful change.

jobrachem:
Thanks @wiep !

PR #372, opened 2026-05-27
Support weak var updates in update_state

jobrachem:
Adds an optional `allow_weak_vars` flag to `Model.update_state` to permit writing position entries for weak variables directly into their cached value nodes. 

The motivation is to allow users to compute gradients with respect to weak variables. I ran into this when working on general autodiff-based working weights for IWLS proposals. For these working weights, the gradient of the log likelihood with respect to a linear predictor is required. A linear predictor is typically weak. This prevented me from obtaining the gradient with autodiff. Hence this PR.

Default behavior remains unchanged (allow_weak_vars=False), there is no backwards-incompatibility. Positions must be unambiguous: no weak variable 
entry may be combined with another entry targeting the same value node, an ancestor, or a descendant. Transient weak variables cannot be set.

Tests cover allowed updates, JAX transform compatibility, ambiguity rejection, and transient weak-var failure cases.

This PR build on #361


jobrachem:
Check: Why does the `model.update()` call at the end of `model.update_state()` not override the state of the weak vars?

jobrachem:
Current behavior is:

1. Override state of weak variables, do not set their flag to `outdated=False`.
2. Update the model state, so descendants of the weak variables are updated.
3. Call `.flag_outdated()` on the value nodes of the updated weak variables.

This way, the returned model state is updated correctly, and a simple `Model.update()` call brings the model back into a consistent state.


Before our meeting today, actually only step 3 was missing. So the update of the state generally worked, but afterwards the weak variables' value nodes were *not* flagged outdated, so a `Model.update()` call would not have had the intended effect.

wiep:
Outdated flags need to propagate through the graph so any node that indirectly depends on the weak variable is also flagged; otherwise the model is inconsistent. Does .flag_outdated() in step 3 handle that propagation?

jobrachem:
Yes, that is the purpose of `.flag_outdated()`. It is an old method, no addition in this PR.

```python
class Node(ABC):
    def flag_outdated(self) -> Node:
        """Flags the node and its recursive outputs as outdated."""
        self._outdated = True

        if self.model:
            for node in self._outputs:
                node.flag_outdated()

        return self
```

wiep:
> Current behavior is:
> 
> 1. Override state of weak variables, do not set their flag to `outdated=False`.
> 2. Update the model state, so descendants of the weak variables are updated.
> 3. Call `.flag_outdated()` on the value nodes of the updated weak variables.

Basically after 2 one reads the value of a descendent of the weak node; at this time the graph is in an inconsistent state but that is the whole point of this operation. Afterwards, 3 brings it back into a consistent state. Sounds good to me as long as there is no thread-safety requirement (which we do not have in liesel).

jobrachem:
>Basically after 2 one reads the value of a descendent of the weak node; at this time the graph is in an inconsistent state but that is the whole point of this operation. Afterwards, 3 brings it back into a consistent state.

Exactly.


>Sounds good to me as long as there is no thread-safety requirement (which we do not have in liesel).

I have not considered threading concerns so far. Are you thinking of a scenario where competing threads enter in a race condition about determining the state of the weak variable in question? If this is the concern, the current behavior guards against such situations, because if a weak variable is in the position dictionary for the update, no ancestor or descendant of that weak variable is allowed in the position. Otherwise, the position would be ambiguous or the update would have to rely on obscure, easy-to-get-wrong mechanism like the order of variables in the position dictionary.

wiep:
we should not consider thread safety.

the liesel model is build and evaluated in one thread. I think that is a reasonable assumption. Once the model is compiled into individual compiled jax functions, those are pure and thread-safe. All threading happens in jax.

src/liesel/model/model.py
```diff
@@ -65,6 +76,18 @@ def fn(at, *args, **kwargs):
     return Calc(fn, var_transformed.value_node, *inputs, **kwinputs)  # type: ignore
 
 
+def _set_weak_var_value(var: Var, value: Array) -> None:
```
wiep, line 79: should we annotate, as a doc string, that this function may put the graph into an inconsistent state and document the contract that `flag_outdated` needs to be called on the variable?

src/liesel/model/model.py
```diff
@@ -2697,11 +2720,47 @@ def extract_position(
 
         return position
 
+    def _node_for_position_key(self, key: str) -> Node:
+        try:
+            return self.nodes[key]
+        except KeyError:
+            return self.vars[key].value_node
+
+    def _validate_weak_var_position(self, position: dict[str, Array]) -> None:
```
wiep, line 2742: docstring:

Raise RuntimeError if any weak variable in `position` is updated alongside
a graph-related nodes such as ancestor, descendant, as such updates are ambiguous.

src/liesel/model/model.py
```diff
@@ -2697,11 +2720,47 @@ def extract_position(
...
+        if not weak_vars:
+            return
+
+        position_nodes = {key: self._node_for_position_key(key) for key in position}
+
+        for weak_key, var in weak_vars:
+            value_node = var.value_node
+            related_nodes = (
+                nx.ancestors(self.node_graph, value_node)
+                | nx.descendants(self.node_graph, value_node)
```
wiep, line 2774: two questions:

- this does not trigger descendant variables that are strong because their value_node is not a descendent of the weak node, correct?
- would it be enough to just check ancestors as decedents and ancestors are symmetric and if descedent B of A is in the update set, then the check on B would be sufficient. Doesn't hurt, though.

wiep (dismissed):
looks solid to me. i would just add those two docstrings to the private methods to document the contract.

src/liesel/model/model.py
jobrachem: 1. Yes, that is correct. I also added an explicit test where a strong variable's distribution depends on a weak variable. Updating both the strong variable and the weak variable is supported, because the strong variable's value node is not a descendant of the weak variable.

2. That would be enough, if we checked the ancestors of all nodes in the update position. Currently, the check only runs with respect to the weak nodes.

src/liesel/model/model.py
jobrachem: Done

src/liesel/model/model.py
jobrachem: Done

wiep (dismissed):

jobrachem:
The approved review was marked as stale, because I just pushed a small formatting update by re-running the pre-commit. Will merge when the tests are finished.

PR #379, opened 2026-06-12
Add support for python 3.14

jobrachem:
Our upper pin on Python <3.14 is no longer necessary, since all dependencies support Python 3.14 and our code does as well. This PR updates the metadata and CI actions accordingly. In the tests, I had to make a tiny change, because the type of exception raised by the Python builtin pickle module changed between 3.13 and 3.14.


jobrachem:
Discussed in a meeting, ok for merge.

wiep:
We may need to bump some minimum dependency versions explicitly for python 3.14 support. With uv's `lowest-direct` resolution, it seems to pick something like scipy 1.14, but compiled wheels for 3.14 only exist from 1.17. I was a bit surprised uv pulled an scipy version without wheels for 3.13 when i was working on #384 and the scipy required version was lower. I don't know if scipy pre 1.14 did not support python 3.13 or there are simply no wheels on pypi. 

We could also run `lowest-direct` resolution only with the oldest supported python version or have python version specific requirements like

```
dependencies = [
  "scipy>=1.14; python_version < '3.14'",
  "scipy>=1.17; python_version >= '3.14'",
]
```
I don't know if that's a good idea, as it complicates the compatibility checks.

jobrachem:
Hm, that is unfortunate. Do I understand correctly that this is mostly a CI issue? Because if a user has Python 3.14, naturally they will use a compatible scipy version.

Python-specific requirements in an action might be annoying to maintain.

wiep:
I think the issue will always occur if a dependency does not provide a tight enough upper bound on the python version. For example, scipy  1.9.3 requires python >= 3.8 (no upper bound) [1] but doesn’t provide wheels for python 3.13. Since the meta data says 3.13 is supported, uv will download the sdist and try to compile. This will happen on the CI as well as on any machine as long as one does not request a newer version of scipy eg with resolution newest (the default).

[1] https://pypi.org/project/scipy/1.9.3/#data

jobrachem:
I see. I think assuming that people will use "newest" resolution in practice except if their setup somehow requires an older version is reasonable. 

What would be your preferred solution? Bumping the dependency or patching the CI action? Between the two, I think I would go for the CI action.

wiep:
we could limit resolution lower direct only on the oldest supported python version. 

If a user wants to use a newer python version but not the newest packages they need to handpick package versions anyway. If they are on a “managed” python the packages will be provided anyhow. 

If a dependency is not working on a newer python version but the dependency does not declare that, I think that is outside of our responsibility.

jobrachem:
Action plan:

- Run resolution-lower-direct (we don't care about the dependencies of the dependencies) only on the oldest supported Python
- Stick with the current behavior of using an upper limit on the supported Python; manual updates should not be so bad.

Issue #381, opened 2026-06-18
Remove Rproj file


wiep:
Liesel no longer has any R dependencies. Consequently, we should be able to remove [liesel.Rproj](https://github.com/liesel-devs/liesel/blob/7141a6b4d019a67c15ed3a06c2f404fa5ee572b5/liesel.Rproj) from this repo.

@jobrachem, any concerns?

jobrachem:
No concerns, I think you are right.

wiep:
Actually, there is a 2nd Rproj file at [docs/source/tutorials/qmd/tutorials.Rproj](https://github.com/liesel-devs/liesel/blob/7141a6b4d019a67c15ed3a06c2f404fa5ee572b5/docs/source/tutorials/qmd/tutorials.Rproj). Do we need to keep that around?

jobrachem:
From my side, that one can go, too. I have not been editing the tutorials in RStudio for a long time now.

PR #384, opened 2026-06-18
Add lowest-direct resolution to weekly cron CI

### LINKED ISSUE #380: Add CI action to verify lower pins of dependencies

wiep:
The weekly cron job now runs two resolution variants

- upgrade (newest deps, as before) and
- lowest-direct (minimum allowed versions).
 
Lower bounds on all direct dependencies were bumped to currently uv.lock versions to avoid resolving to currently not working combinations, including a bump of the pymc lower bound to >=6.0.

Also, the action will now only create an issue if it fails on main.

action passes, see https://github.com/liesel-devs/liesel/actions/runs/27766757019/job/82155505741

closes #380.


jobrachem:
Looks good to me, but I'm wondering if it may be a bit too restrictive to set the lower bounds of all direct dependencies to their most recent current versions?

wiep:
I just took the most recent version in the lock file. I don't think that is the most recent version, unless the lock file has been updated recently.

jobrachem:
Last update was three weeks ago, and I have generally been very liberal with running `uv lock --upgrade` in the past. Chances are that the package versions noted in the lock file are very recent.

wiep:
i see. we can try to lower and see if it passes. pymc needs to be at 6.0 or higher because of changes in pytensor and numpy 2.0, as far as i understand. we could try lock file version from one year ago or would you rather experiment with going as low as possible?

jobrachem:
We can try with an older lock file. In general, I think it would be nice to go as low as possible, that keeps our compatibility broad.

wiep:
ok, i will revert the bumps and use an older version. i need to update scipy and matplotlib because older versions do not support python 3.13. I also tried to use the pytest arg `-W default` so that warnings are treated as warnings and not errors (our config) when using resolution lowest-direct.

However, that makes at least one other test fail. The tests

tests/model/test_node.py::test_frozen_distribution_distribution_manipulation[Dist]
tests/model/test_node.py::test_frozen_distribution_distribution_manipulation[TransientDist]

rely on promoting warnings to errors but tensorflow raises a warning not an error.

```
RuntimeWarning: divide by zero encountered in log 
    lambda x, name=None: np.log(_convert_to_tensor(x)))
```
Either the test should be changed such that it does not rely on the pytest configuration or we should not use  the pytest arg `-W default` when running with lowest-direct resolution. The latter would require to bump tqdm to 4.67 because before it uses a python std lib function marked as deprecated with v3.13. Also, seaborn needs to be bumped from 0.13.0 to 0.13.2.

wiep:
I chose the option to bump tqdm and seaborn. cron job passes. see https://github.com/liesel-devs/liesel/actions/runs/27822149476

added the unrelated removal of two Rproj files to close #381.

jobrachem:
> However, that makes at least one other test fail. The tests
> 
> tests/model/test_node.py::test_frozen_distribution_distribution_manipulation[Dist] tests/model/test_node.py::test_frozen_distribution_distribution_manipulation[TransientDist]
> 
> rely on promoting warnings to errors but tensorflow raises a warning not an error.
> 
> ```
> RuntimeWarning: divide by zero encountered in log 
>     lambda x, name=None: np.log(_convert_to_tensor(x)))
> ```
> 
> Either the test should be changed such that it does not rely on the pytest configuration or we should not use the pytest arg `-W default` when running with lowest-direct resolution. The latter would require to bump tqdm to 4.67 because before it uses a python std lib function marked as deprecated with v3.13. Also, seaborn needs to be bumped from 0.13.0 to 0.13.2.

That's a good point. I think the test should probably use `pytest.warn` instead of `pytest.raises`.

wiep:
> That's a good point. I think the test should probably use pytest.warn instead of pytest.raises.

addressed.

I would allow warnings with resolution lowest-direct, so this passes even when packages deprecate an API and would only fail once that API is removed.

Should we keep the bump of tqdm and seaborn? I'd keep the new tqdm as it would otherwise would produce hundreds of warnings and it is from nov 2024. seaborn v0.13.2 is from jan 2024, so a bump might be ok too but affects less tests so I could revert to v0.13.0.

jobrachem:
> > That's a good point. I think the test should probably use pytest.warn instead of pytest.raises.
> 
> addressed.
> 

Great! :)

> I would allow warnings with resolution lowest-direct, so this passes even when packages deprecate an API and would only fail once that API is removed.
> 
> Should we keep the bump of tqdm and seaborn? I'd keep the new tqdm as it would otherwise would produce hundreds of warnings and it is from nov 2024. seaborn v0.13.2 is from jan 2024, so a bump might be ok too but affects less tests so I could revert to v0.13.0.

I'd be ok with bumping these too, since the bumped versions are not even that recent.

wiep:
Good. I reactivated the warning tolerance on resolution lowest-direct. [The cron action](https://github.com/liesel-devs/liesel/actions/runs/27830114852/) is successful, PR should be ready to be merged.

jobrachem (approved):
Great, thank you @wiep!

PR #385, opened 2026-07-23
Add liesel.optim

jobrachem:
This PR includes comprehensive new optimization functionality for Liesel.

Main highlights:

1. Data splitting (train/validation/test) for Liesel models, allowing for different splitting axes and axis sizes, so that many non-standard models can be used right away.
2. Batching for Liesel models, also allowing for different batching axes and axis sizes.

Main limitations:

1. Both splitting and batching require individual access to log likelihood and log prior for scaling operations. 
    a) For splitting, this is because train, validation, and test likelihoods are brought to the same scale (the training scale) for better comparability.
    b) For batching, this is because batched likelihoods in batched loss evaluations have to be rescaled with respect to the prior to avoid the prior from dominating the likelihood simply because we are using batching.
2. For this reason, the current `NegLogProbLoss` assumes working with a Liesel model. Generalizing towards arbitrary ModelInterface objects would not be that hard, because the `OptimEngine` and the general `Loss` protocol are model-agnostic in principle. But the interfaces would have to differentiate between likelihood and prior.
3. The liesel-focused convenience initializer `LieselOptim` assumes being fed a Liesel model for default setup, it uses `Model.observed`, `Model.parameters`, and `Model.state`.


jobrachem:
In the last weeks, I've been keeping the `obtim-base` branch up to date with the `optim-devel` branch (which includes VI functionality) that I use for my current projects. Recent changes include:

1. Improved progress printing
2. Added the argument `min_epochs` to `Stopper`, which activates an "embargo" period: Optimization will run for at least `min_epochs`, and only after that will early stopping start.
3. Added the option for delaying the start of individual optimizers.

jobrachem:
## Sequence diagram 1: `OptimEngine.fit()`

This really just captures different optimizer branches for different ways of printing progress. 

```mermaid
sequenceDiagram
    autonumber
    actor User
    participant Engine as OptimEngine
    participant Loss
    participant Carry as OptimCarry
    participant Optimizer as Optimizers
    participant JAX as JAX control flow
    participant Progress as tqdm
    participant Result as OptimResult

    User->>+Engine: fit()
    Engine->>Engine: Validate configuration
    Engine->>+Loss: position(optimizer keys)
    Loss-->>-Engine: Initial parameter position

    Engine->>+Carry: new(position, initial state, batches)
    loop Each optimizer
        Carry->>+Optimizer: init(position)
        Optimizer-->>-Carry: Optimizer state
    end
    Carry-->>-Engine: Initialized carry

    alt Progress disabled
        Engine->>+JAX: Run monolithic epoch loop
        loop While stopper permits
            JAX->>Engine: _run_epoch(carry)
            Engine-->>JAX: Updated carry
        end
        JAX-->>-Engine: Final carry
    else Step progress enabled
        loop Epochs and configured batch chunks
            Engine->>JAX: Run jitted batch chunk
            JAX-->>Engine: Carry and status tuple
            Engine->>Progress: Update batch or epoch display
        end
    else Epoch progress enabled
        loop Configured epoch chunks
            Engine->>JAX: Run jitted epoch chunk
            JAX-->>Engine: Carry and status tuple
            Engine->>Progress: Update epoch display
        end
    else Single progress update
        Engine->>JAX: Run monolithic epoch loop
        JAX-->>Engine: Final carry
        Engine->>Progress: Render final status
    end

    Engine->>Engine: Process NaN diagnostics
    Engine->>Engine: Prune unused history
    Engine->>Engine: Select best or final position
    Engine->>Result: Create result
    Result-->>Engine: OptimResult
    Engine-->>-User: OptimResult
```

jobrachem:
## Sequence diagram 2: One batched epoch

One thing we could seriously consider here:

- Currently, the loss is evaluated *after* each step to update the training loss in the carry object.
- We could instead use `jax.value_and_grad` in the `Optimizer.step` method and use that to update the carry training loss.
- This would change the meaning from "after update" to "before update", but it would not affect correctness.
- And it would pretty much cut the number of loss evaluations in half.

```mermaid
sequenceDiagram
    autonumber
    participant Control as JAX epoch loop
    participant Engine as OptimEngine
    participant Batch as Batches / BatchManager
    participant Optimizer
    participant Loss
    participant Transform as Optax transformation
    participant Stopper

    Control->>+Engine: _run_epoch(carry)
    Engine->>+Batch: start_epoch(random key)
    Batch-->>-Engine: Shuffled or resampled batches
    Engine->>Engine: Reset accumulated losses

    loop Each full batch
        Engine->>Batch: get_batched_position(split.train, index)
        Batch-->>Engine: Observed batch

        loop Each optimizer
            alt Activation epoch reached
                Engine->>Optimizer: position(current position)
                Optimizer-->>Engine: Owned parameters
                Engine->>Optimizer: not_position(current position)
                Optimizer-->>Engine: Fixed parameters

                Engine->>+Optimizer: step(parameters, loss, carry)
                Optimizer->>+Loss: grad(parameters, carry)
                Loss->>Loss: Differentiate loss_train_batched
                Loss-->>-Optimizer: Gradient
                Optimizer->>Transform: update and apply_updates
                Transform-->>Optimizer: Updated parameters and state
                Optimizer-->>-Engine: Updated carry.position
            else Optimizer still delayed
                Engine->>Engine: Preserve carry
            end
        end

        Engine->>+Loss: loss_train_batched(updated position, carry)
        Loss-->>-Engine: Post-update batch loss
        Engine->>Engine: Accumulate epoch loss
    end

    alt Validation split exists
        Engine->>+Loss: loss_validate(current position, carry)
        Loss-->>-Engine: Validation loss
    else Exact full-data monitor requested
        Engine->>+Loss: loss_train(current position, carry)
        Loss-->>-Engine: Full training loss
    else Cheap training monitor
        Engine->>Engine: Reuse epoch or weighted average
    end

    Engine->>Engine: Record loss and position history
    Engine->>Engine: Update global best position
    Engine->>Engine: Increment epoch
    Engine-->>-Control: Updated carry

    Control->>+Stopper: continue_(epoch, monitoring history)
    Stopper-->>-Control: Continue or stop
```

jobrachem:
## Sequence diagram 3: One `NegLogProbLoss` evaluation

It may be a bit surprising that the batch and split objects are so important here, but they own relevant information for scaling the likelihood.

```mermaid
sequenceDiagram
    autonumber
    participant Caller as Optimizer or Engine
    participant Loss as NegLogProbLoss
    participant Model as lsl.Model
    participant Scale as Batches or PositionSplit
    participant Graph as Liesel model state

    Caller->>+Loss: Evaluate loss(params, carry)

    alt Mini-batch training
        Loss->>Loss: Merge params, batch, and fixed parameters
        Note over Loss,Scale: Scaling comes from carry.batches
    else Exact full-data training
        Loss->>Loss: Merge params, split.train, and fixed parameters
        Note over Loss,Scale: Scaling comes from the training split
    else Validation
        Loss->>Loss: Merge params, validation data, and fixed parameters
        Note over Loss,Scale: Scaling comes from the validation split
    end

    Loss->>+Model: update_state(position, carry.model_state)
    Model->>Graph: Recalculate dependent nodes
    Graph-->>Model: Updated NodeState mapping
    Model-->>-Loss: Updated model state

    Loss->>+Scale: scaled_log_lik(model, state)

    alt Single likelihood scale
        Scale->>Model: Find observed likelihood nodes
        Model-->>Scale: Distribution-node names
        Scale->>Graph: Sum and scale likelihood values
    else Per-branch scaling
        loop Each batch or split branch
            Scale->>Model: Map observed keys to likelihood nodes
            Scale->>Graph: Sum with branch-specific scale
        end
    end

    Scale-->>-Loss: Scaled log likelihood

    alt Prior included
        Loss->>Graph: Read _model_log_prior.value
        Graph-->>Loss: Log prior
        Loss->>Loss: Combine likelihood and prior
    else Likelihood-only validation
        Loss->>Loss: Use likelihood only
    end

    Loss-->>-Caller: Negative scaled loss
```

jobrachem:
## Sequence diagram 4: `LieselOptim` configuration and engine construction

```mermaid
sequenceDiagram
    autonumber
    actor User
    participant Quick as LieselOptim
    participant Model as lsl.Model
    participant Split as PositionSplit / Manager
    participant Loss as NegLogProbLoss
    participant Batch as Batches / BatchManager
    participant Optimizer as Optimizer / LBFGS
    participant Engine as OptimEngine

    User->>+Quick: __init__(model, options)

    alt Custom loss supplied
        Quick->>Quick: Use loss.split
    else Explicit split supplied
        Quick->>Quick: Use supplied split
    else Derive split
        Quick->>+Split: from_model(model, split options)
        Split->>Model: Read observed variables
        Split->>Model: extract_position(observed keys)
        Model-->>Split: Observed position
        Split-->>-Quick: PositionSplit or PositionSplitManager
    end

    alt Custom loss supplied
        Quick->>Quick: Keep supplied loss
    else Default loss
        Quick->>Loss: Create with model and split
    end

    alt Explicit batches supplied
        Quick->>Quick: Keep supplied batches
    else Derive batches
        Quick->>Batch: from_split(split, batch size)
        Batch-->>Quick: Batches or BatchManager
    end

    alt Explicit optimizer sequence
        Quick->>Quick: Keep supplied optimizers
    else Adam or LBFGS default
        Quick->>Model: Read parameters
        Model-->>Quick: Parameter keys
        Quick->>Optimizer: Create optimizer for all parameters
    end

    Quick-->>-User: Configured LieselOptim

    User->>+Quick: build_engine()
    Quick->>Model: Read model.state
    Model-->>Quick: Initial model state
    Quick->>Engine: Create with loss, batches, optimizers, state
    Engine-->>Quick: Engine
    Quick-->>-User: OptimEngine
```

wiep:
do we discuss this in the meeting or need to be review independently of that?

jobrachem:
We can discuss it in the next meeting I think. Maybe then we decide in favor of further independent review, but for now let's treat this as meeting preparation.

jobrachem:
@wiep @GianmarcoCallegher After our meeting last week, I implemented the following changes:

1. For loss monitoring and early stopping, users now have these three options:
    1. Evaluate loss once on full training data at the end of each epoch. Might be useful for smaller-scale problems that do not use batching and optimize using L-BFGS.
    2. Evaluate loss once on full validation data at the end of each epoch. The classic machine learning setup.
    3. Use an exponentially moving average (EMA) of the history of batch training losses. Probably what you want for VI.
2. Users *always have to* explicitly choose their monitoring loss; there's no magic clever default.
3. Updated the API for the "best" position. 
    1. Dropped the name `OptimResult.best_postion`. Instead we now have `OptimResult.position`, `OptimResult.position_final` and `OptimResult.position_min_monitor`.
    2. `OptimResult.position` is intended as the "continuing position", i.e. the position you continue to work with after optimization. It has the following meaning:
        1. `.position = .position_final` if the monitoring loss is EMA. This is because I interpret EMA's role to be to detect when improvement has reached a plateu, and due to the smoothing window, the "best" EMA loss cannot really be unambiguously paired up with a "best" position.
        2. `.position = .position_min_monitor` if the monitoring loss is validation or full training data. This is because validation monitoring detects when overfitting begins, and full data training monitoring can be clearly paired up with a position that has performed "best".
4. Improved the efficiency of monitoring loss evaluation.
    1. When EMA is the monitoring loss, we now use the pre-update batch losses obtained from `jax.value_and_grad` to avoid a second loss evaluation after each update.
    2. When  validation or full training data monitoring are used, they are computed once at the end of each epoch.


This notebook demonstrates the EMA monitoring: https://github.com/liesel-devs/liesel/blob/optim-base/tests/optim/showcase_ema_monitor.ipynb

jobrachem:
image

I will provide more details on why exactly the EMA looks like this here

jobrachem:
- maybe we would want to include grouping information for the split/batch-managers on liesel variables in an inference object
- for the manager pattern, we could require a grouping via nested lists/tuples

```python
manager = opt.SplitManager.from_model(
    model,
    position_keys=[["y_a"], ["y_b"]], # grouping
    validate_axis_share=0.20,
    test_axis_share=0.10,
)
```

jobrachem:
Diagnosed and fixed https://github.com/liesel-devs/liesel/pull/385#issuecomment-5620433120

Now the EMA loss for the example notebook looks like this:

image

The issue was a computation of the EMA that was insufficiently guarded against float32 rounding. The EMA normalization weight stopped changing at some point (due to many batches per epoch combined with a long monitoring) while the accumulated loss kept growing.

jobrachem:
Also, this branch now includes

1. Checkpointing.
2. Supplying sampling weights for batching.


## **Checkpointing**

```python
result = engine.fit(checkpoint="optim.pkl", checkpoint_every=10)
```

This starts a run if the file doesn’t exist, or resumes from it if it does.

You can also pause to inspect progress:

```python
result = engine.fit(pause_after=100)
result.plot_loss()
result = engine.fit(checkpoint=result.checkpoint)
```

## **Sampling weights for batches**

Allows you to make some observations appear more often in batches, while correctly scaling their likelihood contributions. For example, give observations in sparsely populated intervals a higher chance of being sampled:

```python
import liesel.optim as opt

train_y = split.train["y"]
weights = opt.Batches.weights_binned(train_y, bins=20, strength=0.5)

batches = opt.Batches(
    ["y"],
    axis_size=len(train_y),
    batch_size=64,
    sample_with_replacement=True,
    sampling_weights=weights,
)
```

Using custom weights always requires `sample_with_replacement=True`.

This divides the observed range into 20 equal-width intervals and gives observations in less populated intervals higher weights. `strength=0` gives uniform sampling; `strength=1` gives each occupied interval equal total sampling probability. The default of `0.5` is a middle ground.

`NegLogProbLoss` adjusts the weighted contributions so sampling more often does not give those observations more influence on the overall fit. Weights must follow the training data’s row order.

### Meaning of "strength"

`strength` controls how strongly sampling favors rare bins:

- 0: All observations are equally likely to be sampled.
- 0.5 (default): Partially balances bins using weight = 1 / sqrt(bin count).
- 1: All occupied bins have equal total sampling probability.

General formula: `weight[i] = 1 / (count[i] ** strength)`,
where `count[i]` is the number of observations in observation i’s bin.

### Two other helpers cover categorical data:

```python
# Give each category equal total sampling probability.
weights = opt.Batches.weights_balanced(train_labels)

# Choose the expected share of draws from each category.
weights = opt.Batches.weights_for_shares(
    train_labels, shares={"common": 0.7, "rare": 0.3}
)
```

PR #391, opened 2026-08-26
Update brittle memory chunking tests

jobrachem:
This PR updates two tests that assert chunking lowers compiled temporary memory for `predict` and `sample`, and remove now-unused `_compile_sampling`/`_SamplingSpec` imports. 

These checks relied on the jax function [memory_analysis](https://github.com/jax-ml/jax/blob/2d66622450e2c8633cda2307688ef7aa294bd6eb/jax/_src/stages.py#L183-L184), which, as the docstring states, "need not be consistent across versions of JAX and jaxlib, or even across invocations". The failure was revealed through our pytest cron job action: https://github.com/liesel-devs/liesel/actions/runs/32831224043, which ran with a newer jax version than the locked environment. 

Essentially, the tested computation was so simple that asserting the `chunked memory < unchunked memory` was not consistently true, because the operations were fused very efficiently. The update now tests Cholesky decompositions. They are less easily fused and should provide a more stable test.

### DIFF
--- a/tests/model/test_model.py
+++ b/tests/model/test_model.py
@@ -16,8 +16,6 @@ from liesel.model.model import (
     GraphBuilder,
     Model,
     _compile_prediction,
-    _compile_sampling,
-    _SamplingSpec,
     log_prob_pointwise,
     save_model,
 )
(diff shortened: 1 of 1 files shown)


jobrachem:
We might consider removing the tests in the future, if they still prove to be brittle, since the memory inspection functionality is known to be unreliable.

PR #394, opened 2026-08-28
Use integer Graphviz node IDs for layout

jobrachem:
Convert graph node labels to integers before calling pydot_layout and remap the returned positions to the original node labels. This avoids issues when Graphviz can't handle non-integer/non-safe node identifiers, for example when a node name includes `:`.


wiep:
i requested copilot review out of curiosity. not necessarily something we need to pay attention to.

jobrachem:
> i requested copilot review out of curiosity. not necessarily something we need to pay attention to.

Cool, thanks! I checked what the AI review flagged; it is a false positive. But accepting the proposed explicit cast to int does not hurt and could potentially provide additional robustness. So I'm accepting it.

PR #404, opened 2026-09-19
IWLS and MALA terminology clarification

jobrachem:
This PR clarifies the terminology and behavior of Goose’s IWLS kernel and adds two Langevin kernels.

Previously, `IWLSKernel` used a simplified manifold MALA proposal. For current position `beta`, posterior score `g`, local precision `P`, and step size `s`, its proposal was:

```text
Previous IWLSKernel / new SMMALAKernel:

    mean = beta + (s² / 2) * P⁻¹ g
    covariance = s² * P⁻¹

New IWLSKernel:

    mean = beta + P⁻¹ g
    covariance = P⁻¹
```

No single step size makes the previous proposal match IWLS: `s = 1` matches the covariance but moves the mean only halfway toward the IWLS mean; `s = sqrt(2)` matches the mean but doubles the covariance. With the exact precision of a Gaussian full conditional, the new IWLS proposal gives a Gibbs draw up to numerical error.

**Previously drawn samples remain valid.** The old kernel applied the correct Metropolis-Hastings accept/reject step for its actual proposal. The mismatch concerned terminology and proposal behavior, not the validity of the sampler.

The resulting API changes are:

- **`IWLSKernel`** uses the full IWLS proposal without step-size scaling or adaptation. `.untuned()` remains available and produces the same proposal. Removed tuning arguments raise `TypeError`, and options after `position_keys` and `chol_info_fn` are keyword-only.
- **`SMMALAKernel`** provides the previous algorithm. To preserve the old default behavior, use `SMMALAKernel(..., initial_step_size=0.01)` and retain any other explicit settings. For the old fixed-step proposal, use `SMMALAKernel.untuned(...)`.
- **`MALAKernel`** adds ordinary MALA, using the Langevin proposal above with `P = I` and requiring only first derivatives.
- Both Langevin kernels live in `liesel.goose.mala`, are exported from `liesel.goose`, and select their initial step automatically by default. Their acceptance targets are 0.574 for MALA and 0.8 for SMMALA; explicit initial steps bypass the search.

This is a breaking change. Existing saved sampling results remain loadable for summaries; resuming serialized engines across versions is not covered.

See `docs/source/migration-0.6.md` for migration examples, including how to preserve the previous behavior.


jobrachem:
- There might be slightly different formulations of the IWLS proposal in the literature. We should be sure to get our terminology straight in that regard. @jobrachem will look into this and ask Thomas for input.
- We might still want to avoid the breaking change of the IWLS kernel. An option could be to give the "corrected" kernel a unique name and keep the existing implementation exactly where it is.
- We might also want to reconsider whether the "corrected" IWLS kernel should actually have no option fur step size tuning.

jobrachem:
Alternative Names:

- `LocalGaussianKernel`
- `ScoringKernel`
- `NewtonKernel` / `FisherKernel` to emphasize the relationship to Newton/Fisher scoring

